PR #5, opened 2022-12-15
[WIP] Implementing a more complete version of the starry core matrices

dfm:
- [x] `sT` solution vector
- [x] `A` basis transformation matrices
- [ ] `R` rotation matrices

PR #17, opened 2023-03-01
starry: Rotation matrices

lgrcia:
- Implementation of rotation matrix `R` (#5 )
- Main todo is to replace `@jax.vmap` by `@partial(jnp.vectorize...` with correct signature


dfm:
Awesome - thanks @lgrcia!!

Can you rebase onto `main`? I've merged a few PRs earlier today. Then I'll take a look through and do a review.

Pinging @benjaminpope since this is relevant to other conversations!

dfm:
Edit: never mind - I just changed the base branch here and that worked! You don't need to rebase. I'll take a look tomorrow.

tests/experimental/starry/rotation_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def D(l, alpha, beta, gamma):
+        res = sm.zeros(2 * l + 1, 2 * l + 1)
+        for m in range(-l, l + 1):
+            for n in range(-l, l + 1):
+                res[m + l, n + l] = Dmn(l, m, n, alpha, beta, gamma)
+        return res
+
+    def Umn(l, m, n):
+        """Compute the (m, n) term of the transformation matrix from complex to real Ylms."""
```
dfm, line 61: This line is causing the pre-commit checks to fail because it's too long. Can you break it into two lines?

tests/experimental/starry/rotation_test.py
```diff
@@ -0,0 +1,161 @@
+import numpy as np
+import pytest
+from jax import config
+from jaxoplanet._src.experimental.starry.rotation import R
+
+config.update("jax_enable_x64", True)
```
dfm, line 6: We don't want to add this here! We run the tests with and without x64 enabled because we want to support both approaches. If one of the tests will fail at single precision, we can add a conditional skip to it, but it's better to test both cases if we can.

tests/experimental/starry/rotation_test.py
```diff
@@ -0,0 +1,161 @@
...
+        gamma = sm.atan2(singamma, cosgamma)
+
+        return REuler(l, alpha, beta, gamma)
+
+    def R(lmax, u, theta):
+        """Return the full axis-angle rotation matrix up to degree `lmax`."""
+        blocks = [RAxisAngle(l, *u, theta) for l in range(lmax + 1)]
+        return sm.BlockDiagMatrix(*blocks)
+
+    return R(lmax, u, theta)
```
dfm, line 161: I think it would be useful to add a "compare to starry" test as well, like the ones here:

https://github.com/exoplanet-dev/jaxoplanet/blob/b6417bfc6c3ed6012e9eee6ec97bef2aca43b207/tests/experimental/starry/basis_test.py#L25

and here:

https://github.com/exoplanet-dev/jaxoplanet/blob/b6417bfc6c3ed6012e9eee6ec97bef2aca43b207/tests/experimental/starry/solution_test.py#L41

src/jaxoplanet/_src/experimental/starry/rotation.py
```diff
@@ -0,0 +1,105 @@
...
+                * factorial(l + m - k)
+                * factorial(l - mp - k)
+                * factorial(mp - m + k)
+            )
+        )
+
+        dlm = jnp.nansum(dlm, 0)
+        Dlm = jnp.exp(-1j * (mp * alpha + m * gamma)) * dlm
+
+        return jnp.real(jnp.linalg.inv(U) @ Dlm @ U)
```
dfm, line 90: We probably want to use a `solve` instead of an `inv` here.

dfm:
A more general question: Do you have thoughts about the relationship between your approach here and the one [used in `starry` via recurrence relations](https://github.com/rodluger/starry/blob/b72dff08588532f96bd072f2f1005e227d8e4ed8/starry/_core/ops/lib/include/wigner.h#L23)? I must admit that I don't have a great sense for the performance implications (in terms of efficiency or accuracy) of this direct approach vs. the one implemented in `starry`. What is your sense of this?

benjaminpope:
Hi @dfm & @lgrcia,

Thanks for looping me in - really exciting stuff. We were as of today starting to tackle the rotation matrices with @shashankdholakia and support from ADACS with @jhu-s.

If we have multiple groups of people tackling this perhaps it's worth opening a Discussion or Projects page about the tasks that need to be accomplished? I don't care about credit and I don't think Shashank minds much either, we just want to get things up and running to do good science!

Cheers,

Ben

lgrcia:
Thanks for this review @dfm! 

About the advantages of the recurrence relations, I have no idea. I started implementing them but preferred this more concise code. For efficiency, I kept in mind we would work at fixed `l_max`. For accuracy however I reckon it is worth giving them a try.

@benjaminpope  and all, happy to discuss further!

src/jaxoplanet/_src/experimental/starry/rotation.py
```diff
@@ -0,0 +1,105 @@
...
+    alpha2 = jnp.arctan2(P12 / norm2, P02 / norm2)
+    beta2 = jnp.arctan2(jnp.sqrt(1 - P22**2), P22)
+    gamma2 = jnp.arctan2(P21 / norm1, -P20 / norm1)
+
+    case1 = jnp.logical_and((P22 < -1 + tol), (P22 > -1 - tol))
+    case2 = jnp.logical_and((P22 < 1 + tol), (P22 > 1 - tol))
+
+    alpha = jnp.where(case1, alpha0, jnp.where(case2, alpha1, alpha2))
+    beta = jnp.where(case1, beta0, jnp.where(case2, beta1, beta2))
+    gamma = jnp.where(case1, gamma0, jnp.where(case2, gamma1, gamma2))
```
dfm, line 45: Around these lines I think we'll want to add a test that the gradients don't return `nan`. If the gradient of one branch of the `where` is `nan`, even if it's not used, the `where`'s gradient becomes `nan`.

dfm:
Thanks @lgrcia and @benjaminpope!

I'd say that the approach I'd like to take here is to try to address the inline comments, then merge. That way we have a starting point that @benjaminpope, @shashankdholakia, and @JHu-s can build from. I'd say let's not worry too much about performance for now, but I think it is important to look into our choice of algorithm here in the long run. I'll open some issues to track ideas for next steps.

dfm:
Here's a [discussion thread](https://github.com/exoplanet-dev/jaxoplanet/discussions/23) and [project board](https://github.com/orgs/exoplanet-dev/projects/3) that we can use for discussing next steps. Let's all comment on the thread or open issues as we plan our next steps!

tests/experimental/starry/rotation_test.py
lgrcia: I guess we need to adapt the tolerance of `np.testing.assert_allclose` for single precision tests. Should I write two set of separate tests? (did not see any example of that being done in current tests)

src/jaxoplanet/_src/experimental/starry/rotation.py
lgrcia: I see. Should that be tested in a global test (like `~jnp.isnan(grad(axis_to_euler))`? Or in-line tests in the function to ensure a finite return? (first time with jax autodiff features)

tests/experimental/starry/rotation_test.py
dfm: It already exists here: https://github.com/exoplanet-dev/jaxoplanet/blob/ae01ec962741d1ef08f1b296987ca5dcf8da24a7/src/jaxoplanet/test_utils.py#L5-L11

src/jaxoplanet/_src/experimental/starry/rotation.py
dfm: I was thinking a test case - inline tests wouldn't really do the trick. Choose some inputs that you know will result in NaNs for some branches and test that the gradients are finite!

lgrcia:
I think I am done with these comments! @dfm, let met know if you think about more.

lgrcia:
I just added failing tests for the `dotR` comparison with starry (when rotation axis vector is not unitary). It is rather a misunderstanding of how rotation is applied so I need to work on this. I will update ASAP.

lgrcia:
@dfm, the issue I am left with is that the rotation matrix from starry is not equal to the symbolic one (using sympy) for rotations about the `u=(0, 1, 0)` vector. Mine does, so tests against sympy are passing but not against starry.

The difference between the starry and sympy rotation matrices is only in signs, so it may be because of conventions (?). From all base unit vectors, only the rotation about `(0, 1, 0)` doesn't match, but does for two opposite angles `theta` and `-theta`, if that helps.

There is no Op from starry to get the full rotation matrix so I use `dotR` as:
```python
from starry._core.core import OpsYlm

# map degree, rotation vector and angle
l_max = 3; u = (0, 1, 0); theta=0.1

m = OpsYlm(l_max, 0, 0, 1)

I = np.eye(l_max**2 + 2*l_max + 1)
starry_R = m.dotR(I, *u, theta)
```

The symbolic implementation of the rotation matrix is [from starry](https://github.com/rodluger/starry/blob/v0.2.2/docs/proofs/R.ipynb) (like in [test_rotation.py](https://github.com/lgrcia/jaxoplanet/blob/starry/tests/experimental/starry/rotation_test.py)) and I attach a notebook of the test between starry and sympy (both required): [test_starry_R.ipynb.zip](https://github.com/exoplanet-dev/jaxoplanet/files/10924789/test_starry_R.ipynb.zip)

Any idea?

lgrcia:
@dfm, very straightforward I think! I agree it should match starry. I will try to make informed changes and be sure I understand (and document) starry conventions. Very close to the end I hope. Looking forward sharing more contributions!

dfm:
@lgrcia — I made some changes on main that I've merged into your branch - please pull before continuing!

lgrcia:
Solved! Issue was that Collado 1989 expresses the Dlm matrix as $d_{m' m}$ and I (and Rodrigo's symbolic version) computed $d_{mm'}$, so only a transpose away from matching with starry 🎉 (and I did check that the orientation of the rotations about all basis vectors are as expected).

I will write a `dotR` (for now we have a `Rdot`), test it against `starry` and I think we are done.

lgrcia:
`Rdot` and testings against starry all done  ✨

@dfm, happy with the current PR or you would like to see other things?

Higher level tests will probably be against starry [`YlmBase.rotate`](https://github.com/rodluger/starry/blob/b72dff08588532f96bd072f2f1005e227d8e4ed8/starry/maps.py#L1420), using `dotR` to rotate around different basis vectors a map with a given inclination and obliquity. For this, a proper `Map`-like class needs to be implemented (I am testing one including some neat and fast starry-like visualization features!). These developments are all linked to the general interface we want jaxoplanet to have for bodies, orbits and maps, so might be worse discussing all of that separately, your call @dfm! :)

src/jaxoplanet/_src/experimental/starry/utils.py
```diff
@@ -0,0 +1,72 @@
+import numpy as np
```
dfm, line 1: This module never gets imported - perhaps we should remove it from this PR?

src/jaxoplanet/_src/experimental/starry/utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+from jaxoplanet._src.experimental.starry.basis import Y
```
dfm, line 3: This import won't work (`Y` has been renamed to `Ylm` by you :D) so should be fixed or removed.

src/jaxoplanet/_src/experimental/starry/rotation.py
```diff
@@ -0,0 +1,141 @@
...
+import numpy as np
+from jax.scipy.linalg import block_diag
+from scipy.special import factorial
+
+from jaxoplanet._src.types import Array
+
+
+@jax.jit
+def axis_to_euler(u1: float, u2: float, u3: float, theta: float):
+    """Axis-angle rotation matrix"""
```
dfm, line 15: Could you expand this docstring a bit? I know that the other `starry` matrices don't have good docstrings, but let's start with this one and we can update the others in the next PR.

src/jaxoplanet/_src/experimental/starry/rotation.py
```diff
@@ -0,0 +1,141 @@
...
+    @partial(jnp.vectorize, signature=f"()->({n_max},{n_max})")
+    def _R(theta: Array) -> Array:
+        alpha, beta, gamma = axis_to_euler(u[0], u[1], u[2], theta)
+        full = block_diag(*[rl(alpha, beta, gamma) for rl in Rls])
+        return jnp.where(theta != 0, full, jnp.eye(l_max * (l_max + 2) + 1))
+
+    return _R
+
+
+def Rdot(l_max: int, u: Array) -> Callable[[Array], Array]:
```
dfm, line 113: Perhaps add a docstring?

src/jaxoplanet/_src/experimental/starry/rotation.py
```diff
@@ -0,0 +1,141 @@
...
+    @partial(jnp.vectorize, signature=f"({n_max}),()->({n_max})")
+    def R(y: Array, theta: Array) -> Array:
+        yls = jnp.split(y, idxs)
+        alpha, beta, gamma = axis_to_euler(u[0], u[1], u[2], theta)
+        return jnp.hstack([rl(alpha, beta, gamma) @ yl for rl, yl in zip(Rls, yls)])
+
+    return R
+
+
+def dotR(l_max: int, u: Array) -> Callable[[Array], Array]:
```
dfm, line 127: And here.

dfm (changes requested):
Amazing! This is looking great. Thank you @lgrcia!

I added a few more tiny comments but then I'm very happy merge after you've address those. If you'd be up for opening another PR with your initial implementations of the rest of the model, I'd be very happy to discuss plans there.

dfm (approved):

dfm:
@lgrcia — I've actually removed this from the merge queue because I've just noticed that while the entire test suite runs in 5 minutes on the main branch, this brings that up to 20 minutes. I know that we've asked for a lot of tests, but I think we should try to get this runtime down. Can you do some bench marking to see where that extra runtime is coming from and see if you can simplify the tests?

lgrcia:
@dfm, indeed this is mostly from the symbolic comparison. Last commit greatly improves tests runtime on my side, without loosing too much coverage I think.

dfm:
Much better - thank you!!

PR #27, opened 2023-03-11
Adding interface for arbitrary order limb darkening

### LINKED ISSUE #26: Add support for non-limb darkened light curve
This should be possible with the current implementation in `limb_dark.py`:

https://github.com/exoplanet-dev/jaxoplanet/blob/ae01ec962741d1ef08f1b296987ca5dcf8da24a7/src/jaxoplanet/_src/core/limb_dark.py

But it fails here

https://github.com/exoplanet-dev/jaxoplanet/blob/ae01ec962741d1ef08f1b296987ca5dcf8da24a7/src/jaxoplanet/_src/core/limb_dark.py#L15

and

https://github.com/exoplanet-dev/jaxoplanet/blob/ae01ec962741d1ef08f1b296987ca5dcf8da24a7/src/jaxoplanet/_src/core/limb_dark.py#L151

dfm:
Closes #26


dfm:
@soichiro-hattori This now works as follows:

```python
orbit = orbits.TransitOrbit.init(
    period=10, duration=0.1, time_transit=0.0, radius=0.1, impact_param=0.3)
lc = light_curves.LimbDarkLightCurve.init()
```

and you can use this light curve object how you're currently using the `QuadLightCurve` object, except `LimbDarkLightCurve.init` takes any number of limb darkening coefficients!

PR #48, opened 2023-06-14
Adding transit fitting tutorial

soichiro-hattori:
I've put together a simple transit fitting tutorial notebook that loosely mimics the [transit fitting case study tutorial for the exoplanet package](https://gallery.exoplanet.codes/tutorials/transit/).

It's still rough around the edges since there's a bit more text that would be nice to add.
I'm opening this PR for mainly two reasons:

1. What sort of text or figures are missing in this notebook that would be helpful, and
2. More importantly, **how can we make the sampling faster?** 

To elaborate more on the slow sampling issue, it currently takes ~20 minutes on my laptop to perform NUTS with two parallel chains, each with 500 warmup steps and 1000 subsequent samples. 20 minutes seems a bit slow given the speedup we've been seeing for other cases, so I wanted to see if I was doing anything wrong here.

I know that computing the light curve, i.e., running the line `LimbDarkLightCurve.init(u).light_curve(orbit, t)` where `orbit` is a `TransitOrbit` object, takes ~3 ms per call which seems expensive.

@ssagear


soichiro-hattori:
Not sure whether it's ready for a merge to the main branch since there's still some text I should add, but since the previous commit I've updated

1. Changing the ``regularize_mass_matrix=True`` (default setting ) for the NUTS sampler in ``NumPyro`` to ``regularize_mass_matrix=False``. This greatly reduces the computational time required for sampling (now down to ~3-4 minutes with 4 parallel chains for 500 warmup steps and 2000 production steps compared to ~20 minutes in the previous commit). I haven't done a close investigation but it seems like the posterior itself doesn't change drastically regardless of this change, but the downside is that now we see some divergences in the sampling (28 in this case).

2. Added an additional section at the end that looks at the divergences a bit more closely. It looks like the divergences are clustered around when the impact parameter ``b`` and radius ratio ``r`` are "large". These two parameters also have a nonlinear correlation which is perhaps related to this issue since it indicates curvature in the posterior?

soichiro-hattori:
@dfm I think the issue with the Read the Docs build failing was actually related to the `index.md` file not containing the document name and raising the `WARNING: document isn't included in any toctree` warning (which makes the build fail since `fail-on-warning` is true)?

soichiro-hattori:
@dfm I added `ipywidgets` to the `pyproject.toml` file like you mentioned in the comments of PR #107 and it seemed to fix the progress bar rendering issue!

dfm:
@soichiro-hattori — Awesome!!

dfm:
I think it's time to merge this, and then merge the `tutorials` branch into `main` - what do you think, @soichiro-hattori? We can refactor some of the more general pieces into a separate tutorial later, I think.

PR #61, opened 2023-08-27
Introducing units and refactoring Keplerian orbit models

dfm:
@lgrcia @soichiro-hattori: I've started sketching this out after our conversation the other day.

In this PR, I'm adding support for explicitly specifying units using the [jpu](https://github.com/dfm/jpu) package. It includes some astro/exoplanet specific constants and units, but also adds a decorator (`expect_units`) that should simplify our approach throughout. 

As an example, I've added explicit units to the `KeplerianCentral` class. This will break all the tests for now, but the basic idea is that you can build that class using something like:

```python
from jaxoplanet.units import unit_registry as ureg

central = KeplerianCentral(mass=1.0, radius=1e5*ureg.km)
```

and all the units will be handled appropriately. Then, for example, `central.mass` will always be guaranteed to be a `Quantity` with units `M_sun`.


dfm:
@soichiro-hattori — I want to note while I remember: This PR will break your current use of `TransitOrbit` because of how it handled multi-planet systems. I'd like to change `TransitOrbit` to always only operate on a _single_-planet system because I'm starting to re-think how we want to handle multi-planet systems in general. Let's talk more soon!

dfm:
@lgrcia — Inspired by our conversation today here's a sketch of a version of the orbit interface that supports:

```python
sys = KeplerianSystem(
    Central(radius=0.8),
    bodies=(
        Body(period=10),
    )
)
```

or

```python
sys = KeplerianSystem(Central(radius=0.8))
sys = sys.add_body(Body(period=10.0))
```

or

```python
sys = KeplerianSystem(Central(radius=0.8))
sys = sys.add_body(period=10.0)
```

or, even,

```python
sys = KeplerianSystem(Central(radius=0.8)).add_body(period=10.0).add_body(period=25.0)
```

src/jaxoplanet/orbits/keplerian.py
```diff
@@ -641,71 +604,90 @@ def _get_position_and_velocity(
...
+        # If all the bodies have matching Pytree structure then we save a
+        # stacked version that we can use for vmaps below. This allows for more
+        # efficient evaluations in the case of multiple bodies.
+        self._body_stack = None
+        if len(bodies):
+            spec = list(map(jax.tree_util.tree_structure, bodies))
+            if spec.count(spec[0]) == len(spec):
+                self._body_stack = jax.tree_util.tree_map(
+                    lambda *x: jnp.stack(x, axis=0), *bodies
+                )
```
dfm, line 639: @lgrcia — You might find this interesting... I've managed to work out a way to have the interface that you suggested (which I like very much!), while also getting the good performance of being able to `vmap` over bodies. There are some interesting edge cases where the `vmap` wouldn't work (for example, if some bodies have eccentricity defined, but others don't), and we transparently fall back on a potentially less performant, but general method.

dfm:
I'm going to merge this now because I think it's in good enough shape for now. My next job is docs, but I want to get this in first!

lgrcia:
It looks indeed really good, thanks for the notice! I will try that very soon

PR #63, opened 2023-08-31
Alternative implementation for rotation matrices

dfm:
@lgrcia — This includes a direct port of @rodluger's C++ implementation of the Wigner matrices to JAX. ~~There must be a bug somewhere because all the tests with `u2` are failing, but it's close.~~ _Edit: The bug has been found and fixed._ I find that for `l_max = 10` it's a factor of 16 faster on my computer than the current implementation (see below) and I found a few bugs in your version, so perhaps this can help speed things up for your experiments.

Benchmark on my laptop:

```python
In []: func_old = jax.jit(lambda u, M, theta: dotR(l_max, u)(M, theta))

In []: %timeit func_old(u, M, theta).block_until_ready()
446 µs ± 3.4 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In []: func_new = jax.jit(lambda u, M, theta: dot_rotation_matrix(l_max, *u, theta)(M))

In []: %timeit func_new(u, M, theta).block_until_ready()
27.6 µs ± 597 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Take a look and tell me what you think!


src/jaxoplanet/experimental/starry/rotation.py
```diff
@@ -201,12 +201,10 @@ def dotR(l_max: int, u: Array) -> Callable[[Array], Array]:
         - theta is the rotation angle in radians
     """
     Rls = [Rl(l) for l in range(l_max + 1)]
-    n_max = l_max**2 + 2 * l_max + 1
 
-    @partial(jnp.vectorize, signature=f"({n_max},{n_max}),()->({n_max},{n_max})")
     def R(M: Array, theta: Array) -> Array:
         alpha, beta, gamma = axis_to_euler(u[0], u[1], u[2], theta)
-        return np.hstack(
+        return jnp.hstack(
```
dfm, line 207: There was a regular numpy call which meant that this function couldn't be wrapped in a `jit`. @lgrcia

src/jaxoplanet/experimental/starry/rotation.py
```diff
@@ -201,12 +201,10 @@ def dotR(l_max: int, u: Array) -> Callable[[Array], Array]:
         - theta is the rotation angle in radians
     """
     Rls = [Rl(l) for l in range(l_max + 1)]
-    n_max = l_max**2 + 2 * l_max + 1
 
-    @partial(jnp.vectorize, signature=f"({n_max},{n_max}),()->({n_max},{n_max})")
```
dfm, line 206: @lgrcia — This call to `vectorize` was wrong and (I think) unnecessary. I think it's possible that _all_ the `vectorize` calls are similarly incorrect. Let's investigate.

dfm:
For reference, here's the same benchmark for using starry:

<details><summary>Definitions</summary>

```python
In []: import starry

In []: import theano

In []: import theano.tensor as tt

In []: x = tt.dscalar()

In []: y = tt.dscalar()

In []: z = tt.dscalar()

In []: M_ = tt.dmatrix()

In []: theta_ = tt.dscalar()

In []: func_starry = theano.function([x, y, z, theta_, M_], starry._core.core.OpsYlm(l_max, 0, 0, 1).dotR(M_, x, y, z, theta_))
```
</details>

```python
In []: %timeit func_starry(*u, theta, M)
22.4 µs ± 187 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

lgrcia:
These look absolutely great! Thanks 🙏🏼

dfm:
@lgrcia — I've also just added an experimental implementation of the closed form results for the first few terms of the solution vector, as well as their closed form JVP rules. I was able to refactor the conditionals to get the runtime pretty close to the numerical version (on my computer) and the gradient evaluation if > and order of magnitude more efficient in my tests. I haven't experimented with the GPU yet, and there are some edge cases I haven't accounted for yet... maybe I should move this to a different PR!

dfm:
I've moved the solution vector work over to #70, and I'm going to try to get this properly merged.

dfm:
@lgrcia — I think this is ready to go! My proposal is to remove `rotation.py` because I think this one does everything we need. What do you think about that?

PR #70, opened 2023-10-10
Closed form solution vector

dfm:
I started this implementation in #63, but I think it's better to be separate, so I've reverted the changes there and we'll handle it here.

This will include a closed form implementation of the solution vector, instead of the numerical integral currently implemented. I expect we'll want to keep both implementations because I don't expect this to have very good performance on a GPU, but we should do some experiments!


dfm:
Closing in favor of #71

PR #71, opened 2023-10-10
Implementing closed form solution vector

dfm:


lgrcia:
I will close that for now as this is undeveloped and not in our priorities

PR #72, opened 2023-10-10
Update the basis computations to use sparse linear algebra

dfm:
The basis matrices are all sparse so we can take advantage of that. This PR updates the `A` computation to use `scipy.sparse` (it's all precomputed so it doesn't depend on JAX) and it now produces a [`jax.experimental.sparse.BCOO`](https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.sparse.BCOO.html#jax.experimental.sparse.BCOO) object. The performance doesn't really matter too much since it is computed at compile time, but this will help with memory usage for higher order models.


src/jaxoplanet/experimental/starry/flux.py
```diff
@@ -0,0 +1,74 @@
...
+
+    x = dot_rotation_matrix(
+        deg, -jnp.cos(obl), -jnp.sin(obl), 0.0, -(0.5 * jnp.pi - inc)
+    )(x)
+    x = dot_rotation_matrix(deg, None, None, 1.0, obl)(x)
+    x = dot_rotation_matrix(deg, 1.0, 0.0, 0.0, -0.5 * jnp.pi)(x)
+    x = dot_rotation_matrix(deg, None, None, 1.0, theta)(x)
+    x = dot_rotation_matrix(deg, 1.0, 0.0, 0.0, 0.5 * jnp.pi)(x)
+
+    return x @ y
```
dfm, line 30: @lgrcia — I've added this function to compute the flux which I think you've also looked into. I've found that in some initial benchmarks that this is about and order of magnitude _faster_ than starry on my computer! Let's chat.

PR #73, opened 2023-10-11
Ensuring the square of the kite area calculation is non-negative

soichiro-hattori:
The current implementation of the light curve calculation runs into some issues when the radius ratio `ror > 2.0`. Although the light curve can be generated, the gradients of it (i.e., flux) with respect to `ror` become `NaN` values, causing optimizations and use within the NumPyro framework to break. 

Since the issue was that the square of the kite area could become negative (due to `a+(b-c)` becoming negative when `ror > 2.0`), we've opted to convert all negative values to 0 by taking the maximum between 0 and the calculated value.


dfm:
🚀 Thanks!!

PR #75, opened 2023-10-17
Adding spherical harmonic map interface

dfm:
Based on conversations this morning, here's a sketch of an interface for building spherical harmonic maps in a jax-friendly way. The basic interface is something like:

```python
ylm_map = Ylm(
  {
    (0, 0): 1.0,
    (1, 0): 1.5,
    (5, -2): 0.5,
  }
)
```

which would be equivalent to

```python
starry_map = starry.Map(5)
starry_map[1, 0] = 1.5
starry_map[5, -2] = 0.5
```

Note that there is currently a difference in how `(0, 0)` is treated (I think that our best bet would be to match the starry behavior).

The main thing that I'd want to add before merging would be to implement the product operator. I don't _think_ there's any reason to use the "filter" interface from the original starry implementation vs just updating the map. Therefore, limb darkening would be implemented as the product of two spherical harmonic maps.

cc @lgrcia @benjaminpope @shashankdholakia

### DIFF
--- a/src/jaxoplanet/experimental/starry/wigner3j.py
+++ b/src/jaxoplanet/experimental/starry/wigner3j.py
@@ -1,0 +1,395 @@
+"""The code in this submodule is taken directly from the MIT-licensed
+'spherical' package. Below is the original license text:
+
+Copyright (c) 2021 Mike Boyle
+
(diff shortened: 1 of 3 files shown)


dfm:
Superseded by #111

PR #77, opened 2023-10-18
Implementing render functions

dfm:
After this PR, the following code:

```python
import jax.numpy as jnp
from jaxoplanet.experimental.starry.render import render

y = jnp.array(
    [1.00,  0.22,  0.19,  0.11,  0.11,  0.07,  -0.11, 0.00,  -0.05,
     0.12,  0.16,  -0.05, 0.06,  0.12,  0.05,  -0.10, 0.04,  -0.02,
     0.01,  0.10,  0.08,  0.15,  0.13,  -0.11, -0.07, -0.14, 0.06,
     -0.19, -0.02, 0.07,  -0.02, 0.07,  -0.01, -0.07, 0.04,  0.00]
)
img = render(5, 500, 0.0, 0.0, 0.0, y)
plt.imshow(img, origin="lower")
plt.savefig("render.png", dpi=100, bbox_inches="tight");
```

produces this image:

render

which we can compare to [the starry docs](https://starry.readthedocs.io/en/latest/notebooks/Basics/):

### DIFF
--- a/src/jaxoplanet/experimental/starry/render.py
+++ b/src/jaxoplanet/experimental/starry/render.py
@@ -1,0 +1,42 @@
+import jax.numpy as jnp
+
+from jaxoplanet.experimental.starry.basis import A1, poly_basis
+from jaxoplanet.experimental.starry.wigner import dot_rotation_matrix
+
+
+def ortho_grid(res):
+    x, y = jnp.meshgrid(jnp.linspace(-1, 1, res), jnp.linspace(-1, 1, res))
+    z = jnp.sqrt(1 - x**2 - y**2)
+    y = y + 0.0 * z  # propagate nans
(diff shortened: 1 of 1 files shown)


shashankdholakia:
Sorry for lurking--been using this render function and I noticed a possible typo in left_project that makes this function fail for rotation angles different from 0:

https://github.com/exoplanet-dev/jaxoplanet/blob/361397ce8988eee01d84a1cdec47ca50c74ea971/src/jaxoplanet/experimental/starry/render.py#L19-L34

Last line should return MT, not M.T?

dfm:
That seems like it would be better, doesn't it! Would you be willing to open a small PR to that branch?

dfm:
Superseded by #111

PR #82, opened 2023-11-09
docs: docstrings for few basis functions

lgrcia:
Adding some documentation for some of the basis functions (this is a draft).

For context I needed to express the limb-darkening basis into the polynomial one so these were useful.


dfm:
@lgrcia I think we can close this in favor of #111, right?

Issue #80, opened 2023-11-08
A single Spherical Harmonics rotation using quaternions


lgrcia:
We discussed about using a single rotation of the spherical harmonics instead of combining axis-angle rotations (see [here](https://github.com/exoplanet-dev/jaxoplanet/blob/5e365497ddc8b149f782b5d357090a9a1bdeb12a/src/jaxoplanet/experimental/starry/flux.py#L22)). So here is a post to discuss that.

The only way to do that seems to be using [quaternions](https://www.youtube.com/watch?v=zjMuIxRvygQ), and a robust implementation of Wigner rotation matrices using them. This happen to be described in [Boyle 2013](https://arxiv.org/abs/1302.2919) and implemented in the [spherical](https://github.com/moble/spherical/tree/main) Python/numba package.

I guess we need to re-implement that in jax to get the JIT performances. Given that only the rotation angle about the rotation axis will change I wonder if a single rotation will be optimal. Anyway this implementation is presented as more stable and correct for higher orders (no benchmark?). With that available the only thing left will be to correctly represent the complete rotation with a quaternion.

Any thoughts before I dive into the implementation?

lgrcia:
Another solution is to use the current implementation of the rotation matrices but to compute the axis-angle of the total rotation. Using `sympy`:
```python
from sympy import symbols, Quaternion
from sympy import *

inc, obl, theta = symbols('inc, obl, theta')

q = Quaternion.from_axis_angle((-cos(obl), -sin(obl), 0.0), -(0.5 * pi - inc))
q = q * Quaternion.from_axis_angle((0, 0, 1.0), obl)
q = q * Quaternion.from_axis_angle((1.0, 0.0, 0.0), -0.5 * pi)
q = q * Quaternion.from_axis_angle((0, 0, 1.0), theta)
q = q * Quaternion.from_axis_angle((1.0, 0.0, 0.0), 0.5 * pi)

axis, angle = q.to_axis_angle()
```

Leading to:
```python
def axis_angle(inc, obl, theta):

    numerator1 = np.sqrt(2)/2 * (-np.sin(inc/2) * np.cos((obl - theta)/2) + np.cos(inc/2) * np.cos((obl + theta)/2))
    numerator2 = np.sqrt(2)/2 * (-np.sin(inc/2) * np.sin((obl - theta)/2) + np.sin((obl + theta)/2) * np.cos(inc/2))
[...]

PR #83, opened 2023-11-11
Fused rotation operator

dfm:
In our meeting @lgrcia and I implemented #80 and it works!

We wanted to try splitting it into 3 rotations instead of just one though. @lgrcia feel free to push to this branch!


dfm:
Superseded by #111

Issue #84, opened 2023-11-13
Error creating LimbDarkLightCurve due to jpu units?


shashankdholakia:
I'm getting an error when trying to compute a limb darkened light curve with the latest dev version of jaxoplanet. I think this is related to the introduction of jax units. Here is an MWE to reproduce the error:

```python
from jaxoplanet.light_curves import LimbDarkLightCurve
from jaxoplanet import light_curves, orbits
import jax.numpy as jnp
import matplotlib.pyplot as plt

#Fiducial planet parameters:
params = {'period':300.456,
          'radius':0.1,
          'u':jnp.array([0.3,0.2]),
          'bo':0.8,
}
# The light curve calculation requires an orbit
orbit = orbits.keplerian.Body(period=params['period'], radius=params['radius'], impact_param=params['bo'])

# Compute a limb-darkened light curve using jaxoplanet
t = jnp.linspace(-0.3, 0.3, 1000)

lc = LimbDarkLightCurve(params['u']).light_curve(orbit, t=t)

plt.plot(t, lc[0], color="C0", lw=2)
plt.ylabel("relative flux")
plt.xlabel("time [days]")
_ = plt.xlim(t.min(), t.max())
```

which returns: `TypeError: Unexpected input type for array: `

soichiro-hattori:
Hi @shashankdholakia! Thanks for bringing this issue up. There was a small bug where we weren't calling the correct functions. We should have been calling `jpu.numpy` functions instead of `jnp.numpy` functions when using the arrays with units (e.g., `t`). The changes made in PR #86 should fix this issue but let us know if it still doesn't!

One small change is that the `light_curve` function now returns a 1D array with size `len(t)` so you won't need to index the first element of `lc` when plotting the flux (`lc[0] -> lc`)!

shashankdholakia:
Thanks, this fixed it!

PR #86, opened 2023-11-13
Fixing issue where light curve computation fails because of incorporating JPU units

soichiro-hattori:
This PR addresses the issue raised in #84 where the `LimbDarkLightCurve` class's `light_curve` function raises an error because of a TypeError. This error is caused by passing `pint.Quantity` objects into `jnp` functions even though a JAX array is expected. Since `pint.Quantity` objects can be correctly handled by the `jpu.numpy` functions, substituting the current `jax.numpy` function calls with the corresponding `jpu.numpy` function calls when these objects are being passed fixes this issue.

Other small changes were made so that the exposure time integration still works with the units. There is one hacky part, `dt = jnp.outer(texp.magnitude, dt) * texp.units`, where since `jnpu` does not support the `outer` function yet I'm manually separating the magnitude and the unit of `texp`, doing the computation using the usual `jnp.outer` function, and then reattaching the unit. @dfm: Maybe we could implement the outer function in `jpu.numpy`? 

The returned light curves are now 1-D vectors so indexing is not necessary (unless you actually want the first flux calculation).


dfm:
Thanks @soichiro-hattori! Can you add a test to `tests/light_curves_test.py` (it would be fine to just copy the example from #84 minus the plotting)?

> Maybe we could implement the outer function in jpu.numpy?

Good idea - I'm actually surprised that it's not already there! Can you open an issue on that repo?

> The returned light curves are now 1-D vectors so indexing is not necessary (unless you actually want the first flux calculation).

This is actually a little subtle - it depends on the type of orbit! But I think that's a conversation for another thread.

soichiro-hattori:
> Thanks @soichiro-hattori! Can you add a test to `tests/light_curves_test.py` (it would be fine to just copy the example from #84 minus the plotting)?

The updated commit should have the test in there!
> 
> > Maybe we could implement the outer function in jpu.numpy?
> 
> Good idea - I'm actually surprised that it's not already there! Can you open an issue on that repo?

Will do!

soichiro-hattori:
Since I've added a PR on the `jpu` repo to add the `outer` function, I've updated this commit to use the `jnpu.outer` function instead of the hacky approach mentioned above.

dfm:
Looks great to me - thanks!!

PR #74, opened 2023-10-12
Implementing starry "flux" function

dfm:

PR #89, opened 2023-11-21
feat: starry limb-darkened flux

lgrcia:
A `flux` implementation for non-uniform limb-darkened maps. This merges lots of other branches so I open the PR mainly for discussion and I am happy to open another one once the mergings are figured out (See #74).

As discussed with @dfm, the limd-darkening map product is done on the polynomial representation. This has several advantages:
- Map product implementation is simpler (see `Pijk` products compared to`Ylm`)
- The rotation is done on maps with lower degree (and not on the full `u_deg + y_deg` degree)
- Once `y` is represented as `p_y`, in the polynomial basis , the design matrix for `p_y` is `x = jnp.where(b_occ, sTA2, rT(full_deg))` (less matrices multiplications?)

This version has been tested against _starry_ (see `tests/flux_test.py`). The sparse matrices still have to be figured out to avoid calls to `todense` (can you help with that @dfm?). Also tests in float32 don't pass... (half the light curve is nan, to be solved)

Edit: I also added the `theta_z` rotation in the fused rotation, very simple edit and computing the `left_project` is straight-forward (see `rotation.py`). I hope we can use the `None` args for efficiency, as was implemented for the `dot_rotation_matrix`.


src/jaxoplanet/experimental/starry/flux.py
```diff
@@ -0,0 +1,79 @@
...
+    U = jnp.array([1, *u])
+    full_deg = deg + u_deg
+    b = jnp.sqrt(jnp.square(xo) + jnp.square(yo))
+    b_rot = jnp.logical_or(jnp.greater_equal(b, 1.0 + ro), jnp.less_equal(zo, 0.0))
+    b_occ = jnp.logical_not(b_rot)
+
+    # Occultation
+    theta_z = jnp.arctan2(xo, yo)
+    sT = solution_vector(full_deg)(b, ro)
+    sTA2 = jnp.linalg.solve(A2_inv(full_deg).T.todense(), sT.T).T
```
dfm, line 22: Here we might be better off using something like:

```suggestion
    A2 = scipy.sparse.linalg.inv(A2_inv(full_deg))
    A2 = jax.experimental.sparse.BCOO.from_scipy_sparse(A2)
    sTA2 = sT @ A2
```

(Note: I haven't actually tested this, but it does seem like `A2` is sparse.)

src/jaxoplanet/experimental/starry/flux.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # full rotation
+    rotated_y = left_project(deg, inc, obl, theta, theta_z, y)
+
+    # limb darkening product
+    p_y = Pijk.from_dense(A1(deg).todense() @ rotated_y, degree=deg)
```
dfm, line 28: Similarly, here we could use:

```suggestion
    A2_val = jax.experimental.sparse.BCOO.from_scipy_sparse(A1(deg))
    p_y = Pijk.from_dense(A2_val @ rotated_y, degree=deg)
```

src/jaxoplanet/experimental/starry/flux.py
```diff
@@ -0,0 +1,79 @@
...
+    p_u = Pijk.from_dense(U @ U0(len(u), deg), degree=u_deg)
+    p_y = p_y * p_u
+
+    x = jnp.where(b_occ, sTA2, rT(full_deg))
+    norm = np.pi / (rT(u_deg) @ p_u.todense())
```
dfm, line 33:
```suggestion
    norm = np.pi / (p_u.tosparse() @ rT(u_deg))
```

Maybe?

src/jaxoplanet/experimental/starry/flux.py
```diff
@@ -0,0 +1,79 @@
...
+
+    return (x @ p_y.todense()) * norm
```
dfm, line 35: Same here?

```suggestion
    return (p_y.tosparse() @ x) * norm
```

dfm (commented):
I left some tiny comments about how to avoid the `todense` calls. I'm not sure how necessary these are, but it's an idea!

dfm:
Superseded by #111

PR #92, opened 2023-12-06
Adding Zenodo info for releases

dfm:
@lgrcia — What's your ORCID?


lgrcia:
My ORCID is [0000-0002-4296-2246](https://orcid.org/0000-0002-4296-2246)

PR #107, opened 2024-02-08
docs: synthetic radial velocities

lgrcia:
A simple tutorial to compute and fit the radial velocities of a star hosting an exoplanet


dfm:
Thanks! Take a look at the rendered version here: https://jaxoplanet--107.org.readthedocs.build/en/107/tutorials/rv/

Some thoughts:

1. I don't love that we see `_body_stack` in the printed system - I will open a separate PR about that!
2. Could you add a semicolon at the end of plotting cells so that we don't get output like:
```
<ErrorbarContainer object of 3 artists>
```
3. The progress bar output from sampling is pretty unwieldy - perhaps it would be enough to install `ipywidgets` as part of the [`docs` extra in `pyproject.toml`](https://github.com/exoplanet-dev/jaxoplanet/blob/93c413eac76e293da9f192dd74b4b4826c6dca2e/pyproject.toml#L30)?

dfm:
@soichiro-hattori: it looks like the progress bar is still getting muddled on this one. Do you know what's going on? Maybe it's because we need to run parallel sampling to get the notebook-style progress bars?

soichiro-hattori:
That's interesting! I don't have a good idea but I imagine it has to do with what you mentioned about running a single chain vs multiple chains! I "think" that there are different calls to the `tqdm` package based on the number of chains from what's written here https://github.com/pyro-ppl/numpyro/pull/900. 

@lgrcia could you try adding this line to the top of your notebook
`from tqdm import tqdm.notebook as tqdm`

soichiro-hattori:
@lgrcia: I guess that didn't make a difference :( If you add `numpyro.set_host_device_count(2)` right after `import numpyro` you should be able to force the chains to run in parallel and get rid of this progress bar issue.

dfm:
I'm going to merge this now and we can tweak the formatting in a new PR.

Thanks @lgrcia!!

lgrcia:
Thanks 🙏🏼

PR #81, opened 2023-11-08
Simple graticule plot (and visualization module)

lgrcia:
A simple graticule plot function

```python
import matplotlib.pyplot as plt
from jaxoplanet.experimental.starry.visualization import graticule

n = (8, 6)
graticule(0.5, 0.5, n=n, axis=(0, 2), behind=True, c="0.95")
graticule(0.5, 0.5, n=n, axis=(0, 2))

# equal aspect ratio
plt.gca().set_aspect("equal")
plt.savefig("graticule.png", dpi=100, bbox_inches="tight")
```
graticule

PR #111, opened 2024-02-09
feat: interface for starry maps and light curves 

lgrcia:
This is a big PR with all starry features, leading to a `Map` object with a `light_curve` function. This includes:
- fused rotation using sympy expression
  - #80
  - #83
- Spherical harmonics `Ylm` map interface
  - #75
- Polynomial `Pijk` map interface with multiplication
- Limb-darkening coefficients transforms to polynomial basis
- Rendering and map plotting functions
  - #77
  - #81
- `Map` object using the `Ylm` map interface 
- starry light_curve function
  - #74
  - #89
 - many fixes to make all the above work together

Some things to chat about:
- I deleted the wigner3j interface since we don't use it anymore (we multiply using the polynomials)
- Is it properly jax.jitable?
- adding units

For now, here is a self-contained example of what that looks like

```python
from jaxoplanet.experimental.starry import Map, Ylm
import numpy as np
import jax

np.random.seed(30)
y = Ylm.from_dense(np.random.rand(20))
m = Map(y=y, u=[0.5, 0.1], inc=0.9, obl=-0.3)
m.show()
```
Screenshot 2024-02-08 at 10 43 18 PM

```python
from jaxoplanet.orbits.keplerian import System, Central, Body

central = Central()
system = System(central, bodies=[
    Body(radius=0.1, period=1., time_transit=0.1),
    Body(radius=0.2, period=1.)
])

time = np.linspace(-0.2, 0.2, 1000)
lc = m.light_curve(system, time)

plt.plot(time, lc.T)
```
Screenshot 2024-02-08 at 10 43 44 PM


src/jaxoplanet/experimental/starry/pijk.py
```diff
@@ -0,0 +1,140 @@
...
+
+class Pijk(eqx.Module):
+    """A class to represent and manipulate spherical harmonics in the polynomial basis.
+    Several indices are used throughout the class:
+    - Indices (i, j, k) represent the order of the polynomials of (x, y, z), for example
+    (1, 0, 2) represents x * z^2.
+    - Indices (l, m) represent the orders of the spherical harmonics.
+    - Index n represent the index of the polynomial in the flattened array.
+
+    Flattened array `to_dense` and `from_dense` follow th convention from Luger et al. (2019).
```
dfm, line 22: It looks like this line is [tripping up the linter](https://results.pre-commit.ci/run/github/353677214/1707494294.T0l5dHFhQiCykEIjsAvshg) since it is too long.

dfm:
One other random thought: perhaps we shouldn't remove the Wigner stuff, even if we're not using it right now, since it might be useful for future use cases?

src/jaxoplanet/experimental/starry/pijk.py
```diff
@@ -0,0 +1,140 @@
...
+    @classmethod
+    def from_dense(cls, x: Array, degree: int = None) -> "Pijk":
+        data = defaultdict(float)
+        if degree is None:
+            for i in range(len(x)):
+                data[Pijk.n2ijk(i)] = x[i]
+        else:
+            n = (degree + 1) ** 2
+            for i in range(n):
+                data[Pijk.n2ijk(i)] = x[i] if i < len(x) else 0.0
```
dfm, line 109: We probably don't need to loop all the way up to `n`, since all of the higher order entries are `0.0` because of this conditional. We can just loop up to `len(x)`!

lgrcia: True but we then need to set the `degree` of the map and modify the `toparse` function to output an array consistent with that degree. This is important for certain operations in the `maps.Map.render` and `light_curves.light_curve` functions. I think the indices in `toparse` should be computed from the degree anyway instead of taken from `data`. What do you think?

lgrcia:
I fixed the starry light curve tests for the float32 precision ([3f5f0a6](https://github.com/exoplanet-dev/jaxoplanet/pull/111/commits/3f5f0a6e742fe4be09f327078979acda7ff8f62c), now all passing 🎉), the `Pijk` degree thing ([a8089ec](https://github.com/exoplanet-dev/jaxoplanet/pull/111/commits/a8089ec48eea4b1328dd211276d059de4ea69960)) and revived the `wigner3j.py`([359baed](https://github.com/exoplanet-dev/jaxoplanet/pull/111/commits/359baed4a6f7daa9bac68edaba882176422216f8)).

@dfm, do you see anything else before merging?

src/jaxoplanet/experimental/starry/basis.py
```diff
@@ -17,6 +17,17 @@
...
+
+    Parameters
+    ----------
+    lmax : int
+        maximum degree of the spherical harmonic basis
+
+    Returns
+    -------
+    TODO
+    """
```
dfm, line 30: Can you update the docstrings to be in the "Google" format that is used elsewhere?

https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

src/jaxoplanet/experimental/starry/maps.py
```diff
@@ -0,0 +1,128 @@
...
+        kwargs.setdefault("alpha", 0.5)
+
+        # plot lines
+        lat, lon = lon_lat_lines(pts=pts)
+        lat = rotate_lines(lat, self.inc, self.obl, theta)
+        plot_lines(lat, **kwargs)
+        lon = rotate_lines(lon, self.inc, self.obl, theta)
+        plot_lines(lon, **kwargs)
+        theta = np.linspace(0, 2 * np.pi, 2 * pts)
+        plt.plot(np.cos(theta), np.sin(theta), **kwargs)
```
dfm, line 105: I'd probably refactor these visualization methods out into separate functions, because I feel like there's a slight interface inconsistency here. For example, we would want to be able to `jit` the `render` method inside of `show`, or to use `render` within a `jit` (or `vmap`, or `grad`) context, but it doesn't really make sense to `jit` `show`. I think it's worth thinking about what interface makes sense, but I don't know what the answer is - I'm just pretty sure this isn't it :D

src/jaxoplanet/experimental/starry/maps.py
```diff
@@ -0,0 +1,128 @@
...
+                theta,
+            ),
+            signature="(),(),(),()->()",
+        )
+
+        def body_lc(body):
+            x, y, z = body.relative_position(time)
+            return lc_func(x.magnitude, y.magnitude, z.magnitude, body.radius.magnitude)
+
+        lc = jax.vmap(body_lc)(system._body_stack)
```
dfm, line 126: This won't work in general because only a Keplerian `System` has a `_body_stack`. In [`LimbDarkLightCurve`](https://github.com/exoplanet-dev/jaxoplanet/blob/02f55c63536ccce24a3fe8d3c392133f79dc41d1/src/jaxoplanet/light_curves.py#L25), we extract the relevant quantities manually and restrict the allowed input systems using the [`LightCurveOrbit`](https://github.com/exoplanet-dev/jaxoplanet/blob/02f55c63536ccce24a3fe8d3c392133f79dc41d1/src/jaxoplanet/proto.py#L14) protocol, but I don't think that's a good solution.

Probably what we want to do is have a common API for mapping functions over bodies in orbits. This is related to some of the things @soichiro-hattori has been thinking about wrt exposure time integration. Maybe it's ok to leave this as it is, and think about what a better abstraction would be later.

src/jaxoplanet/experimental/starry/pijk.py
```diff
@@ -0,0 +1,139 @@
...
+
+class Pijk(eqx.Module):
+    """A class to represent and manipulate spherical harmonics in the polynomial basis.
+    Several indices are used throughout the class:
+    - Indices (i, j, k) represent the order of the polynomials of (x, y, z), for example
+    (1, 0, 2) represents x * z^2.
+    - Indices (l, m) represent the orders of the spherical harmonics.
+    - Index n represent the index of the polynomial in the flattened array.
+
+    Flattened array `to_dense` and `from_dense` follow th convention from Luger et al.
```
dfm, line 22:
```suggestion
    Flattened array `to_dense` and `from_dense` follow the convention from Luger et al.
```

src/jaxoplanet/experimental/starry/utils.py
```diff
@@ -0,0 +1,91 @@
+import jax.numpy as jnp
+import numpy as np
+from scipy.spatial.transform import Rotation
+
+from jaxoplanet.experimental.starry.basis import A1, poly_basis
+from jaxoplanet.experimental.starry.rotation import left_project
+
+
+def ortho_grid(res):
```
dfm, line 9: I think that `res` needs to be an `int` and it needs to be static at compile time right. Perhaps we should wrap this in

```python
@partial(jax.jit, static_argnums=0)
```

src/jaxoplanet/experimental/starry/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    y = y + 0.0 * z  # propagate nans
+    x = jnp.ravel(x)[None, :]
+    y = jnp.ravel(y)[None, :]
+    z = jnp.ravel(z)[None, :]
+    lat = 0.5 * jnp.pi - jnp.arccos(y)
+    lon = jnp.arctan2(x, z)
+    return (lat, lon), (x, y, z)
+
+
+def render(deg, res, theta, inc, obl, y):
```
dfm, line 21: Like above, we might want to be clear about exactly which values need to be static?

dfm (changes requested):
Looking great! I'm going through a proper code review now and here are some high level thoughts:

1. If I understand correctly, one limitation of the current version is that the `Map` is implicitly applied to the `Central` body in a system. Do you think it would be possible to connect `Map`s to any `Body` instead? How would that change the high level interface?
2. We should export the public API in `.../starry/__init__.py`. Currently everything needs to be imported from the appropriate submodule so the example code doesn't work.
3. It might be useful to add a stub for a tutorial describing this interface with a few examples. We can expand that later, but it might be nice to get it started with this PR?

src/jaxoplanet/experimental/starry/maps.py
lgrcia: I kept this one for a later PR. for now I just deleted the `light_curve` function

dfm (approved):
This all looks good to me! I'd say: let's merge and we can handle any issues in future pull requests?

lgrcia:
I think this PR is ready now. Probably next steps would be:
- Solving the #130
- When creating a body a dummy `Map` is instantiated. Allow map to be `None`
- Handling all the edge cases with `None` values. The ones I think about are:
  - `r = xo = yo = z0 = None` in `map_light_curve`
  - `map = None` in `map_light_curve`
  - `Map(ydeg = 0)`
- Fix and compare the spot expansion from `starry`
- Rename `map` argument everywhere? (`map` is also a python built-in function)

Anything else to keep in mind, @dfm, @soichiro-hattori? 

Apart from that I am happy to merge!

dfm:
I'd say merge away and open new issues to track the next steps. Thanks @lgrcia !!

Issue #112, opened 2024-02-09
LimbDarkLightCurve().light_curve() method fails for multiplanetary systems


soichiro-hattori:
The current version of the `LimbDarkLightcurve().light_curve()` method fails for multiplanetary systems, for example when using the `keplerian` orbit and passing more than one body to the `bodies` argument for `keplerian.System`. 

The following code snippet will produce the error.

```
import numpy as np
import matplotlib.pyplot as plt
import jax
import jax.numpy as jnp
import jpu.numpy as jnpu

from jaxoplanet.orbits import keplerian
from jaxoplanet.light_curves import LimbDarkLightCurve
from jaxoplanet.types import Quantity
from jaxoplanet.units import unit_registry as ureg

jax.config.update("jax_enable_x64", True)

star = keplerian.Central(mass=5, radius=2)
planet_a = keplerian.Body(radius=0.05, period=1)
planet_b = keplerian.Body(radius=0.2, period=2)
system = keplerian.System(central=star, bodies=[planet_a])

t = np.linspace(-5, 5, 500)
lc = LimbDarkLightCurve().light_curve(system, t)
```
with the error:
```
{
	"name": "TypeError",
	"message": "Unexpected input type for array: <class 'pint.Quantity'>",
	"stack": "---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 20
     17 system = keplerian.System(central=star, bodies=[planet_a])
     19 t = np.linspace(-5, 5, 500)
---> 20 lc = LimbDarkLightCurve().light_curve(system, t)

File ~/miniforge3/envs/jaxoplanet/lib/python3.11/site-packages/equinox/_module.py:861, in BoundMethod.__call__(self, *args, **kwargs)
    860 def __call__(self, *args, **kwargs):
--> 861     return self.__func__(self.__self__, *args, **kwargs)

File ~/jaxoplanet/src/jaxoplanet/units/decorator.py:173, in QuantityInput.__call__.<locals>.wrapped(*args, **kwargs)
    165     if unit is not None:
    166         bound_args.arguments[name] = jax.tree_util.tree_map(
    167             partial(_apply_units, name=name, strict=self.strict),
    168             value,
    169             unit,
    170             is_leaf=_is_quantity,
    171         )
--> 173 return func(*bound_args.args, **bound_args.kwargs)

[...]

PR #113, opened 2024-02-09
Making changes to LimbDarkLightCurve().light_curve() method to make for multiplanetary systems

soichiro-hattori:
This PR resolves Issue #112 so that the light_curve() method works for multiplanetary systems. 

I think we should think a bit more about whether we want to squeeze the output array into 1d when there's only one body or to always return an array of shape (N_bodies X N_time) even if it's only one body. 
The first option might be a bit more intuitive but the second option allows us to streamline our code since the more general code of handling more than one body will work even if only one body is passed.


dfm:
Thanks! Can you update this PR to only include the relevant changes? Right now it looks like it also includes the tutorial from #7. It might be easiest to make a new branch from the upstream main and apply your changes there. I'm happy to talk that through in more detail! It would also be useful to include the test from #112.

PR #114, opened 2024-02-09
Making changes to LimbDarkLightCurve().light_curve() method to make for multiplanetary systems

soichiro-hattori:
New PR to replace old PR #113. Changes were made so that these commits are independent from the `tutorials` branch changes.

Same text as PR #113: 
This PR resolves Issue https://github.com/exoplanet-dev/jaxoplanet/issues/112 so that the light_curve() method works for multiplanetary systems.

I think we should think a bit more about whether we want to squeeze the output array into 1d when there's only one body or to always return an array of shape (N_bodies X N_time) even if it's only one body.
The first option might be a bit more intuitive but the second option allows us to streamline our code since the more general code of handling more than one body will work even if only one body is passed.


src/jaxoplanet/light_curves.py
```diff
@@ -66,6 +66,10 @@ def light_curve(
         if texp is None:
             tgrid = t
         else:
+            if (
+                texp.ndim == 0
```
dfm, line 70: Are we sure that `texp` will always have a `.ndim` attribute? What about when a `float` is passed? Maybe we should use `jnp.ndim`?

src/jaxoplanet/light_curves.py
```diff
@@ -66,6 +66,10 @@ def light_curve(
...
+            ):  # If a scalr is passed, automatically broadcast to the shape of t
+                texp = jnpu.repeat(texp, len(t))
```
dfm, line 72: I think `jnpu.broadcast_to(texp, jnp.shape(t))` might be better?

tests/light_curves_test.py
```diff
@@ -21,3 +21,14 @@ def test_light_curve():
...
+
+def test_multiplanetary():
+    star = orbits.keplerian.Central(mass=5, radius=2)
+    planet_a = orbits.keplerian.Body(radius=0.05, period=1)
+    planet_b = orbits.keplerian.Body(radius=0.2, period=2)
+    system = orbits.keplerian.System(central=star, bodies=[planet_a, planet_b])
+
+    t = jnp.linspace(-5, 5, 500)
+    lc = LimbDarkLightCurve().light_curve(system, t)
+    assert lc.shape == (system.shape[0], t.size)
```
dfm, line 34: Should we also test a version with `texp`?

dfm:
I think your point about what we want the semantics of this operation to be for 1 vs. multi-planet systems is a good one!

My feeling is that at a high level we want to support orbits with different "shapes". For example we could have an orbit with shape `()` (a scalar), which just has one planet, or an orbit with shape `(1,)` which also has only one planet. In these cases, I would expect the shapes of the light curve to be `(len(t))` and `(1, len(t))` respectively.

But, that thought really just pushes the question up a level to a discussion of how we handle shapes in multi-planet orbits...

So, basically I have no good answer!

src/jaxoplanet/light_curves.py
soichiro-hattori: I'm pretty sure `texp` will always have an `.ndim` attribute since it's automatically converted to a `pint.Quantity` object. This behavior is true even if `texp` is a `float`.

But I'm happy to change it!

src/jaxoplanet/light_curves.py
soichiro-hattori: Okay!

tests/light_curves_test.py
soichiro-hattori: Yes, I've added 2 more tests with `texp` being passed.

soichiro-hattori:
Okay that's the current behavior anyways! 
I think this PR is ready to be merged. @dfm are there other changes that need to be made before that happens?

dfm (approved):

dfm:
Great - I've sent this to the merge queue! Thank you!!

PR #119, opened 2024-02-10
Drafting faster transit fitting tutorial

dfm:
I'm not sure that we'd want to merge this, but I've sketched out some ideas that make the transit fitting example a bit faster to run. One extra point here is that I'm also not getting any divergences anymore, so we can factor that part out of this tutorial (perhaps we could talk about it in the "intro to numpyro" tutorial?). The only real changes that I made were to:

1. Reduce parallelism (RTDs definitely doesn't allocate 4 cores)
2. Reduce size of dataset smaller and choose somewhat easier parameters

What do you think, @soichiro-hattori?

### DIFF
--- a/docs/tutorials/transit.ipynb
+++ b/docs/tutorials/transit.ipynb
@@ -40,7 +40,7 @@
     "import copy\n",
     "\n",
     "numpyro.set_host_device_count(\n",
-    "    4\n",
+    "    2\n",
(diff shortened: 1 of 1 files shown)


soichiro-hattori:
This is awesome! I've removed text and the section on divergences and I think this is ready to merge.

PR #123, opened 2024-02-16
feat: public `body_vmap` (meeting with @dfm)

lgrcia:
A public `body_vmap` function to vmap over system's bodies (from @dfm)

EDIT: See `tests/orbits/keplerian_test.py` for examples (and tests)

### DIFF
--- a/src/jaxoplanet/__init__.py
+++ b/src/jaxoplanet/__init__.py
@@ -1,2 +1,7 @@
-from jaxoplanet import core as core, orbits as orbits, units as units
+from jaxoplanet import (
+    core as core,
+    light_curves as light_curves,
+    orbits as orbits,
+    units as units,
+)
 from jaxoplanet.jaxoplanet_version import __version__ as __version__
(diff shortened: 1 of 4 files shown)


dfm:
I love this! I just pushed a somewhat major refactor, but don't have time to explain. Take a look when you have a chance and I'll explain my reasoning soon!

dfm:
I've added some more tests and docs to this branch - sorry for changing your implementation so much, @lgrcia!! The changes I made probably over-engineer things a bit, but I wanted to support a wide range of the common simple uses of `vmap`. The main interface change from what you implemented is that the user is no longer responsible for passing the `in_axes` spec for the first argument (the `Body`). The rest of the changes are just to support arbitrary Pytrees of `in_axes` and `out_axes` specifications. This means reproducing some of the logic from `vmap`, but I think it's probably worth it!

@lgrcia — can you take a look through this when you get a chance and let me know what you think?

lgrcia:
This looks amazing, happy to merge!

PR #129, opened 2024-02-21
Experimenting with a different interface for exposure time

dfm:
@soichiro-hattori — Check this out! I think this is potentially cool, and might simplify some of our logic around light curve models. Basically the idea is that we can factor the exposure time integration out of the `LightCurve` object into a function transformation. This is a bit more "JAX-y" and means that we could support different kinds of light curve models more easily. The use would look something like:

```python
import jax
import jax.numpy as jnp

from jaxoplanet.orbits import keplerian
from jaxoplanet.light_curves import LimbDarkLightCurve
from jaxoplanet.exposure_time import integrate_exposure_time


central = keplerian.Central(mass=5, radius=2)
system = keplerian.System(central).add_body(radius=0.05, period=3.0)

def light_curve(time):
    return LimbDarkLightCurve().light_curve(system, time)[0, 0]

integrated_light_curve = integrate_exposure_time(light_curve, exposure_time=0.1, num_samples=15)
t = jnp.linspace(-5, 5, 5000)
lc = jax.vmap(integrated_light_curve)(t)
```

The new `integrate_exposure_time` function operates on any function which takes a scalar time as an argument and calls that function vmapped onto the appropriate time grid to integrate.

What do you think??

### DIFF
--- a/docs/tutorials/transit.ipynb
+++ b/docs/tutorials/transit.ipynb
@@ -27,7 +27,7 @@
    "outputs": [],
    "source": [
     "import jaxoplanet\n",
-    "from jaxoplanet.light_curves import LimbDarkLightCurve\n",
+    "from jaxoplanet.light_curves import limb_dark_light_curve\n",
     "from jaxoplanet.orbits import TransitOrbit\n",
     "import numpy as np\n",
     "import matplotlib.pyplot as plt\n",
@@ -84,7 +84,7 @@
     "# Compute a limb-darkened light curve for this orbit\n",
     "t = np.linspace(-0.1, 0.1, 1000)\n",
     "u = [0.1, 0.06]  # Quadratic limb-darkening coefficients\n",
-    "light_curve = LimbDarkLightCurve(u).light_curve(orbit, t)\n",
+    "light_curve = jax.vmap(limb_dark_light_curve(orbit, u))(t)\n",
     "\n",
     "# Plot the light curve\n",
     "plt.figure(dpi=150)\n",
@@ -123,13 +123,13 @@
     "ROR = 0.08  # planet radius / star radius\n",
     "U = np.array([0.1, 0.06])  # limb darkening coefficients\n",
     "yerr = 5e-4  # flux uncertainty\n",
-    "t = np.arange(0, 17, 0.02)  # day\n",
+    "t = np.arange(0, 17, 0.05)  # day\n",
     "\n",
     "\n",
     "orbit = TransitOrbit(\n",
     "    period=PERIOD, duration=DURATION, time_transit=T0, impact_param=B, radius=ROR\n",
     ")\n",
-    "y_true = LimbDarkLightCurve(U).light_curve(orbit, t)\n",
+    "y_true = jax.vmap(limb_dark_light_curve(orbit, U))(t)\n",
     "y = y_true + yerr * random.normal(size=len(t))\n",
     "\n",
     "# Let's see what the light curve looks like\n",
@@ -190,7 +190,7 @@
     "    orbit = TransitOrbit(\n",
     "        period=period, duration=duration, time_transit=t0, impact_param=b, radius=r\n",
     "    )\n",
-    "    y_pred = LimbDarkLightCurve(u).light_curve(orbit, t)\n",
+    "    y_pred = jax.vmap(limb_dark_light_curve(orbit, u))(t)\n",
     "\n",
     "    # Let's track the light curve\n",
     "    numpyro.deterministic(\"light_curve\", y_pred)\n",
@@ -505,7 +505,7 @@
     "    impact_param=inferred_b,\n",
     "    radius=inferred_r,\n",
     ")\n",
-    "y_model = LimbDarkLightCurve(inferred_u).light_curve(orbit, t)\n",
+    "y_model = jax.vmap(limb_dark_light_curve(orbit, inferred_u))(t)\n",
     "\n",
     "fig, ax = plt.subplots(dpi=150)\n",
     "\n",
@@ -536,6 +536,13 @@
     "ax.legend(fontsize=10, loc=4)\n",
     "ax.set_xlim(-inferred_duration, inferred_duration);"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
   }
  ],
  "metadata": {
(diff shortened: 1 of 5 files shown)


dfm:
@soichiro-hattori — A question for you: In your implementation of exposure time integration you set the default units for `texp` to be seconds rather than days. I'm inclined to use days since that's our default time unit throughout, then a user could specify seconds explicitly if they want. Any opinions on that?

soichiro-hattori:
> The new `integrate_exposure_time` function operates on any function which takes a scalar time as an argument and calls that function vmapped onto the appropriate time grid to integrate.
> 
> What do you think??

I like the more fp/JAX-y style! I'm still trying to wrap (no pun intended) my head around some of the `@wraps`, `__call__`, and `...`, so I'll add more detailed thoughts tomorrow, but to me if we're writing `integrate_exposure_time` as a function I'd expect it to work on an array of flux values of the _unintegrated light curve_ (i.e., the output of calling `LimbDarkLightCurve.light_curve()`). But right now it seems like it's wrapping a function and then returning another function. Maybe this issue can just be solved by another level of abstraction where we define an `LimbDarkLightCurve.integrated_light_curve()` and then calling this machinery?


> @soichiro-hattori — A question for you: In your implementation of exposure time integration you set the default units for `texp` to be seconds rather than days. I'm inclined to use days since that's our default time unit throughout, then a user could specify seconds explicitly if they want. Any opinions on that?

Nope! I think those units were there before I made some changes since I remember also thinking that it might be confusing to a potential user that the two time arguments were being assumed to be different units. Although on the other hand I could see the argument that conventional units for observation times are days while for exposure time it's seconds!

dfm:
Thanks @soichiro-hattori! Good points all around. I have a deadline for another project today, so I don't have time to follow up, but I'll take a stab at clarifying my ideas about this tomorrow!

soichiro-hattori:
@dfm: Here are the changes I've made and some example code to show how it currently works!

- Since the `light_curve` method you implemented only accepted one scalar value of `time` instead of an array, I've renamed it to `_light_curve` and then additionally defined a `light_curve` method that takes care of the vmapping over time. I'd understand if you prefer to have the user explicitly vmap themselves, but I just thought it might be worth handling the vmapping in the back!
```python
    def _light_curve(
        self, orbit: LightCurveOrbit, time: Quantity, *, ld_order: int = 10
    ) -> Quantity:
        return self(orbit, time, ld_order=ld_order)

    def light_curve(
        self, orbit: LightCurveOrbit, time: Quantity, *, ld_order: int = 10
    ) -> Array:
        def _lc_func(time: time) -> _LightCurveFunc[T]:
            return self._light_curve(orbit=orbit, time=time, ld_order=ld_order)

        return jax.vmap(_lc_func, in_axes=(0))(time)
```
- I've used your `integrate_exposure_time` function and called it within the `LimbDarkLightCurve` class to define a `integrated_light_curve` method. I'm not sure if this aligns with the separation you had in mind between the `LightCurve` object and the exposure time integration functionality, but this way allows for the user to instantiate an `LimbDarkLightCurve` object and defer the choice of exposure time integration until needed. I suppose this change makes them back to being coupled a bit, but since the `integrated_light_curve` method is pretty much boilierplate code maybe we can handle it with a base `LightCurve` class? 
```python
    def _integrated_light_curve(
        self,
        orbit: LightCurveOrbit,
        time: Quantity,
        *,
        ld_order: int = 10,
        exposure_time: Optional[Quantity] = None,
        order: int = 0,
        num_samples: int = 7,
    ) -> _LightCurveFunc[T]:
        def _partial_lc_func(time):
            return self._light_curve(orbit=orbit, time=time, ld_order=ld_order)

        _integrated_lc_func = integrate_exposure_time(
            _partial_lc_func, exposure_time=exposure_time, num_samples=num_samples
        )
        return _integrated_lc_func

    def integrated_light_curve(
        self,
        orbit: LightCurveOrbit,
        time: Quantity,
        *,
        ld_order: int = 10,
        exposure_time: Optional[Quantity] = None,
        order: int = 0,
        num_samples: int = 7,
    ) -> Array:
        _integrated_lc_func = self._integrated_light_curve(
            orbit=orbit,
            time=time,
            ld_order=ld_order,
            exposure_time=exposure_time,
            order=order,
            num_samples=num_samples,
        )
        return jax.vmap(_integrated_lc_func, in_axes=(0))(time)
```

Below is some example code and the output you'd get!
```python
import jax
import jax.numpy as jnp
import jpu.numpy as jnpu

from jaxoplanet.orbits import keplerian
from jaxoplanet.light_curves import LimbDarkLightCurve
from jaxoplanet.exposure_time import _LightCurveFunc, integrate_exposure_time

central = keplerian.Central(mass=5, radius=2)
system = keplerian.System(central).add_body(radius=0.05, period=3.0)

ldlc = LimbDarkLightCurve()
t = jnp.linspace(-1, 1, 500)
unintegrated_lc = ldlc.light_curve(system, t)  # No exposure-time integratation
integrated_lc = ldlc.integrated_light_curve(system, t,
                                            exposure_time=0.1,
                                            num_samples=15)  # Exposure-time integrated

plt.plot(t, unintegrated_lc, label='texp=0')
plt.plot(t, integrated_lc, label='texp=0.1d')
plt.legend()
```
image

- Slightly unrelated to this, but the output shape of the `light_curve` function is now `(t.size, N_bodies)` which is the transpose of before. Should we just transpose it after the vmap?
- I'm guessing some of this will break with multiple bodies. I'll test it out tomorrow.

dfm:
@soichiro-hattori Based on our chat, here's an updated syntax:

```python
import jax
import jax.numpy as jnp

from jaxoplanet.orbits import keplerian
from jaxoplanet.light_curves import limb_dark_light_curve
from jaxoplanet import exposure_time

central = keplerian.Central(mass=5, radius=2)
system = keplerian.System(central).add_body(radius=0.05, period=3.0)

ldlc = limb_dark_light_curve(system, 0.2, 0.3)
int_ldlc = exposure_time.integrate(ldlc, exposure_time=0.1, num_samples=15)
t = jnp.linspace(-5, 5, 5000)

plt.plot(t, jax.vmap(ldlc)(t))
plt.plot(t, jax.vmap(int_ldlc)(t))
```

Does that about fit what we decided on?

lgrcia:
@dfm happy to make the changes. The new `main` should be merged on this one right? I think the branch was created before #111 was merged

soichiro-hattori:
@dfm: I think we should merge!

dfm:
🚀

Issue #130, opened 2024-02-21
`body.central.mass` in a system 


lgrcia:
The body central mass is different from the system central mass when defined this way:
```python
from jaxoplanet.orbits import keplerian

central = keplerian.Central(radius=1., mass=1.8)
body = keplerian.Body(radius=1., period=1.0, mass=1.9)
system = keplerian.System(central, bodies=[body])

system.bodies[0].central.mass, system.central.mass
```
```
(1.0 <Unit('M_sun')>, 1.8 <Unit('M_sun')>)
```

However using `add_body` works!
```python
from jaxoplanet.orbits import keplerian

central = keplerian.Central(radius=1., mass=1.8)
system = keplerian.System(central).add_body(radius=1., period=1.0, mass=1.9)

system.bodies[0].central.mass, system.central.mass
```
```
(1.8 <Unit('M_sun')>, 1.8 <Unit('M_sun')>)
```

Should the first way be avoided or can we fix that?

dfm:
Fixed by https://github.com/exoplanet-dev/jaxoplanet/pull/131

PR #131, opened 2024-02-22
Making it harder to instantiate Body objects on their own

dfm:
This is my proposed fix for #130. The basic idea is that we no longer allow a `Body` to be instantiated without an explicit `Central`, encouraging users to use the `System.add_body` method instead. It's still possible to create bodies using:

```python
central = Central(mass=0.9, radius=1.2)
body = Body(central, period=1.5)
```

but

```python
body = Body(period=1.5)
```

now throws an error.


lgrcia:
Given our discussion I totally agree! The API documentation should be quite clear about that (and all tutorials). I'll check them!

lgrcia (approved):

dfm:
So. I took another go at this and I think it's better! I've added a new object called `OrbitalBody` which replaces the old `Body` object and `Body` now just holds the parameters that are specified. This means that the

```python
system = System(Central(mass=0.9, radius=1.1), bodies=[Body(period=1.0)])
```

syntax works again, and if you inspect `system`, you'll see:

```python
System(
  central=Central(
    mass=<Quantity(0.9, 'M_sun')>,
    radius=<Quantity(1.1, 'R_sun')>,
    ...
  ),
  bodies=(
    OrbitalBody(
      central=Central(
        mass=<Quantity(0.9, 'M_sun')>,
        radius=<Quantity(1.1, 'R_sun')>,
        ...
      ),
      period=<Quantity(1.0, 'day')>,
      ...
    ),
  ),
  _body_stack=BodyStack(...)
)
```

dfm:
@lgrcia — For some reason some starry comparison tests are failing as a result of this PR. I can't figure out why for the life of me - could you take a look? Thanks!

src/jaxoplanet/orbits/keplerian.py
```diff
@@ -169,240 +308,90 @@ class Body(eqx.Module):
...
+            self.sin_inclination = jnpu.sin(body.inclination)
             self.impact_param = self.cos_inclination / dcosidb
         else:
-            self.impact_param = jnpu.zeros_like(self.period)
-            self.cos_inclination = jnpu.zeros_like(self.period)
-            self.sin_inclination = jnpu.ones_like(self.period)
+            z = jnp.zeros_like(self.period.magnitude) * ureg.dimensionless
+            self.impact_param = z
+            self.cos_inclination = z
+            self.sin_inclination = z
```
soichiro-hattori, line 380: @dfm: I think `self.sin_inclination` is supposed to be initialized with `1`s instead of `0`s (`z`)!

lgrcia: Nice catch @soichiro-hattori! I wasn't sure you could so I just pushed the changes. I hope it's fine!

dfm: Good catch @soichiro-hattori!! Thanks for tracking that down!

PR #133, opened 2024-02-23
feat: starry spot expansion + maps with amplitude and better default

lgrcia:
- starry spot expansion (tested).
- Add an `intensity` function to the map (use same logic for `render`)
 - Add `__init__(normalize=True)` to maps and `relative` kwarg. `Map.intensity` and `map_light_curve` functions are multiplied by the amplitude
 - Default `Map` of a `Body` has an amplitude of 0. and default `Map` of a `Central` has an amplitude of 1.
 - Default `Ylm` is `Ylm({(0,0): 1.0})` and by defaults the Ylm coeffs are normalized so that `Ylm[(0,0)] == 1.0`.
 - Fix `deg=0` issue in `map_light_curve`.
 - Few more refactoring and docstrings.
 
 For reference https://starry.readthedocs.io/en/latest/notebooks/EclipsingBinary_Generate/ is not immediately reproducible since it uses `Map.add_spot` which is deprecated. I rather implemented the `Map.spot` interface with more relatable arguments (spot radius instead of Gaussian std, and spot contrast instead of resulting flux amplitude).


src/jaxoplanet/orbits/keplerian.py
```diff
@@ -400,7 +400,7 @@ def __init__(
             self.time_transit = jnpu.zeros_like(self.period)
 
         if map is None:
-            self.map = Map(y=Ylm({(0, 0): 0, (1, 0): 0.0}))
+            self.map = Map(amplitude=0.0)
```
dfm, line 403: This is a place where I think it would be good to allow a `None` map to indicate a system where there is no map because then we can speed up the computations rather than computing a light curve and then multiplying by zero!

tests/experimental/starry/light_curve_test.py
```diff
@@ -71,26 +71,32 @@ def test_compare_starry(deg, u):
             "central": keplerian.Central(
                 mass=1.3,
                 radius=1.1,
-                map=Map(y=1.0),
+                map=Map(
+                    y=Ylm.from_dense(np.hstack([1.0, np.random.rand(10) * 0.1])),
```
dfm, line 75: I don't think we should use `random` here. If we must, let's use the following instead:

```suggestion
                    y=Ylm.from_dense(np.hstack([1.0, np.random.default_rng(0).uniform(0.0, 0.1, 10)])),
```

tests/experimental/starry/light_curve_test.py
```diff
@@ -119,24 +125,24 @@ def test_compare_starry_system(keplerian_system):
...
         r=central.radius.magnitude,
         m=central.mass.magnitude,
         prot=central.map.period,
     )
     if central.map.u:
         pri.map[1:] = central.map.u
-
-    pri.map[1:, :] = central.map.y.todense().__array__()[1:]
+    if central.map.deg > 0:
+        pri.map[1:, :] = central.map.y.todense().__array__()[1:]
```
dfm, line 137: Is this call to `__array__` to convert to numpy arrays? I normally use something like `np.asarray` because I think it's clearer...

dfm (changes requested):

src/jaxoplanet/orbits/keplerian.py
lgrcia: I am preparing a separate PR where I handle all the `None` arguments in the `starry.light_curve` module (see [feat/none-args-lightcurves](https://github.com/lgrcia/jaxoplanet/tree/feat/none-args-lightcurves))

dfm: Great!

dfm (approved):

PR #141, opened 2024-02-27
Vectorize light curve functions

dfm:
This PR updates the interface introduced by #129 to remove the requirement that light curves always be manually vmapped. But, it still allows the logic within the light curve computation to assume that it only needs to operate on a scalar time. I think this is the best of both worlds because it makes the implementation of light curves a little more straightforward while keeping a nice user interface that can operate on arrays of times.

To do this, I added a `light_curves.utils.vectorize` decorator which handles the vmapping and units. We can't use [`jnp.vectorize`](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.vectorize.html) (which would have been my preference) because we need to support `Quantity` input and output.


lgrcia:
Looks nice! Once merged I'll try to use that for the starry light curves functions too.

dfm:
I'm going to merge this for now so that we can keep rolling with some other PRs, but I'd love to hear your thoughts later when you have a chance @soichiro-hattori!

soichiro-hattori:
@dfm: This all looks great!

PR #142, opened 2024-02-27
Adding interpolation transform

dfm:
Sometimes it can be useful to precompute the light curve on a grid near a transit and then interpolate those computations to the required phases when computing the full model. This can speed things up a lot when you have many transits, or a lot of out of transit data.

This PR adds a first stab at adding this as a function transform similar to the exposure time integration. The basic interface would be as follows:

```python
from jaxoplanet.light_curves.interpolate import interpolate

# Set up your base light curve as usual
ldlc = limb_dark_light_curve(orbit, u)

# Transform it to an interpolated model
interp_ldlc = interpolate(
    ldlc,
    period=orbit.period,
    time_transit=orbit.time_transit,
    num_samples=200,
    # This should always be longer than the actual transit duration;
    # it sets how much time is included around the transit when
    # pre-computing
    duration=0.2,
)

flux = interp_ldlc(time)  # compute the light curve
```


soichiro-hattori:
@dfm: `jnp.allclose(ldlc(time), flux)` evaluates to `False` even with `x64` enabled. Is this fine? 
Here's a code snippet to reproduce it:

```python
import jax 
import jax.numpy as jnp
import numpy as np

from jaxoplanet.light_curves import limb_dark_light_curve
from jaxoplanet.light_curves.transforms import interpolate
from jaxoplanet.orbits import TransitOrbit

# Simulate some data with Gaussian noise
random = np.random.default_rng(42)
PERIOD = random.uniform(2, 5)  # day
T0 = PERIOD * random.uniform()  # day
DURATION = 0.08  # day
B = 0.5  # impact parameter
ROR = 0.08  # planet radius / star radius
U = np.array([0.1, 0.06])  # limb darkening coefficients
time = np.arange(0, 10, 0.01)  # day


orbit = TransitOrbit(
    period=PERIOD, duration=DURATION, time_transit=T0, impact_param=B, radius=ROR
)

# Set up your base light curve as usual
ldlc = limb_dark_light_curve(orbit, U)

# Transform it to an interpolated model
interp_ldlc = interpolate(
    ldlc,
    period=orbit.period,
    time_transit=orbit.time_transit,
    num_samples=200,
    # This should always be longer than the actual transit duration;
    # it sets how much time is included around the transit when
    # pre-computing
    duration=0.2,
)

flux = interp_ldlc(time)  # compute the light curve

jnp.allclose(ldlc(time), flux)  # Evaluates to False
```

```python
plt.plot(time, ldlc(time))
plt.plot(time, flux, ':')
```
image

```python
plt.plot(ldlc(time)-flux, 'k')
```
image

dfm:
Good point! I think this is typically to be expected. I think that if you continued to increase `num_samples` you would find that eventually it would hit the expected precision.

Adding a test like this, maybe with a high tolerance would be good!

soichiro-hattori:
You're right. I just tried it with `num_samples=10_000` and it evaluated to `True`.

PR #144, opened 2024-02-28
Refactoring surface map interface

### LINKED ISSUE #134: Remove `map` from `Body` and `Central` objects
Mostly just a note to self. I think the interface of specifying a `Map` when creating a `Body` is great, but I think it probably makes sense to have a subclass or wrapper for bodies with surface maps.

@lgrcia

dfm:
Fixes #134 

This is the beginning of a PR to refactor the surface map interface. There are a few changes to make the light curve functions compatible with the changes we made in #129 and #141, but the main difference is that I've introduced a `jaxoplanet.experimental.starry.orbit` module which includes a `SurfaceMapSystem` (name totally up for debate!), providing the main interface for building systems with surface maps. The interface would look something like:

```python
system = SurfaceMapSystem(central, central_map).add_body(period=1.0, surface_map=...)
```

Another change is that `light_curve` returns a tuple with the central light curves and body light curves separately rather than a stack. This allows for better support of `None` maps.

Another nice feature of this is that `body_vmap` and `surface_map_vmap` are now separate functions.

cc @lgrcia, @dgegen


src/jaxoplanet/experimental/starry/light_curves.py
```diff
@@ -76,7 +90,11 @@ def map_light_curve(map, r: float, xo: float, yo: float, zo: float, theta: float
     Returns:
         ArrayLike: light curve
     """
-    U = jnp.array([1, *map.u])
+    if map is None:
+        return jnp.zeros_like(theta)
+
+    # TODO(lgrcia): Is this the right behavior when map.u is None?
+    U = jnp.array([1, *(() if map.u is None else map.u)])
```
dfm, line 97: @lgrcia Can you check this behavior for `map` is `None` and `map.u` is `None`?

src/jaxoplanet/experimental/starry/light_curves.py
```diff
@@ -88,19 +106,22 @@ def map_light_curve(map, r: float, xo: float, yo: float, zo: float, theta: float
...
-    rotated_y = left_project(
-        map.ydeg, map.inc, map.obl, theta, theta_z, map.y.todense()
-    )
+    # TODO(lgrcia): Is this the right behavior when map.y is None?
+    if map.y is None:
+        rotated_y = jnp.zeros(map.ydeg)
+    else:
+        rotated_y = left_project(
+            map.ydeg, map.inc, map.obl, theta, theta_z, map.y.todense()
+        )
```
dfm, line 121: @lgrcia Also here?

dgegen:
### Fixing `starry/light_curve_test.py`
The changes made to `test_compare_starry_system` ensure its successful execution. The main changes are the use of `assert_allclose` to compare fluxes element-wise, a check to avoid comparisons if jaxoplanet's flux is `None`, and the initialisation of a limb-darkened map in `Starry` if the body in `jaxoplanet` has no associated map.

```python
def test_compare_starry_system(keplerian_system):
    def extract_starry_map_kwargs(jaxoplanet_map):
        if jaxoplanet_map is None:
            # starry's simplest case: a limb darkened map
            map_kwargs = dict(ydeg=0, rv=False, reflected=False)
        else:
            map_kwargs = dict(
                ydeg=jaxoplanet_map.ydeg,
                udeg=jaxoplanet_map.udeg,
                inc=np.rad2deg(jaxoplanet_map.inc),
                obl=np.rad2deg(jaxoplanet_map.obl),
                amp=jaxoplanet_map.amplitude,
            )
        return map_kwargs

    starry = pytest.importorskip("starry")
    starry.config.lazy = False

    # jaxoplanet system
    central = keplerian_system.central
    central_map = keplerian_system.surface_map
    body = keplerian_system.bodies[0]
    body_map = keplerian_system.surface_maps[0]

    time = np.linspace(-1.5, 1.0, 300)
    jaxoplanet_flux = light_curve(keplerian_system)(time)

    # starry system
    central_map_kwargs = extract_starry_map_kwargs(central_map)
    pri = starry.Primary(
        starry.Map(**central_map_kwargs),
        r=central.radius.magnitude,
        m=central.mass.magnitude,
        prot=0 if central_map is None else central_map.period,
    )
    if central_map and central_map.u:
        pri.map[1:] = central_map.u
    if central_map and central_map.deg > 0:
        pri.map[1:, :] = np.asarray(central_map.y.todense())[1:]

    body_map_kwargs = extract_starry_map_kwargs(body_map)
    sec = starry.Secondary(
        starry.Map(**body_map_kwargs),
        r=body.radius.magnitude,
        m=body.mass.magnitude,
        porb=body.period.magnitude,
        prot=0 if body_map is None else body_map.period,
    )
    if body_map and body_map.u:
        sec.map[1:] = body_map.u
    if body_map and body_map.deg > 0:
        sec.map[1:, :] = np.asarray(body_map.y.todense())[1:]

    starry_system = starry.System(pri, sec)
    starry_flux = starry_system.flux(time, total=False)

    # compare
    for jaxoplanet_flux_item, starry_flux_item in zip(jaxoplanet_flux, starry_flux):
        if jaxoplanet_flux_item is not None:
            assert_allclose(jax.numpy.squeeze(jaxoplanet_flux_item), starry_flux_item)
```
With these changes, I pass all tests in `starry/light_curve_test.py` on my machine.

dfm:
@dgegen — Awesome! I was actually just in the process of handling this issue. We still need some of your fix though. I updated the interface so that `light_curve` always just returns one array, so you don't need to worry about the `None` in the output. But, we still need the `extract_starry_map_kwargs` function that you added. Could you open a PR to the `fix-134` branch?

src/jaxoplanet/experimental/starry/light_curves.py
```diff
@@ -1,68 +1,88 @@
...
+                system.radius, -xos, -yos, -zos, time
+            )
+
+        result = jnp.zeros(system.shape, dtype=time.dtype)
+        if central_light_curves is not None:
+            result += central_light_curves
+        if body_light_curves is not None:
+            result += body_light_curves
+
+        return result
```
dfm, line 77: My remaining question here is whether or not we'd ever want to expose the central and body light curves separately or does it make sense to always add them together?

dgegen: Let me rephrase the question: _Why should we prevent a curious user from learning about the decomposition of the light curve into the individual contributions of the various bodies in the system?_
Perhaps we could add a keyword to return the sum of all light curves instead of the individual light curves (or vice versa, depending on what kind of default behaviour we think would be the more intuitive).

PS. If I understand you correctly, I should open another PR to pull from `fix-134` in my fork to `fix-134` in this repository.

dfm: > Perhaps we could add a keyword to return the sum of all light curves instead of the individual light curves (or vice versa, depending on what kind of default behaviour we think would be the more intuitive).

I think that's exactly the question!

> PS. If I understand you correctly, I should open another PR to pull from fix-134 in my fork to fix-134 in this repository.

Yes please!

shashankdholakia: Starry's default behavior is to return the total flux by default--I think that would make sense here too

https://starry.readthedocs.io/en/latest/System/#starry.System.flux

dgegen: I agree. Especially since getting an array instead is a more natural result compared to a tupel containing both `None` values and arrays. So I think we should also set something like `total=False` by default.

dfm: @shashankdholakia — The `total` here is actually a little different! In `jaxoplanet` our tradition is to always return separate arrays for each body. The question here is how we should handle the fact that there are separate light curves for the central and orbiting bodies. Do we want to sum the central and the orbiting bodies? Or have an array that has a shape of `1 + n_bodies`.

shashankdholakia: In that case, I'd prefer to have `1+n_bodies` with the central prepended to the front since it could be useful to model phase curves of the planets individually

lgrcia: A `total` keyword is easily replaceable by a `.sum` with a fine control over the axes behind summed (thinking about multi-wavelength maps here). So I think I would personally prefer to see a different behavior than starry here. Although that's only minor

dfm: Great! In that case, I think that the right thing to return would be a tuple with `(central_light_curve, body_light_curves)` since these are qualitatively different objects. `central_light_curve` would have shape `(num_times,)` and `body_light_curves` would have shape `(num_bodies, num_times)`.

Then we can add a keyword argument (like `total`) to decide if we want everything to be summed.

dgegen: What about the case where one of the orbiting bodies has no surface map, i.e. surface maps were only specified for some but not all of the orbiting bodies? Will its whole column in `body_light_curves` just be filled with 0 or `NaN` values?

src/jaxoplanet/experimental/starry/light_curves.py
lgrcia: I think the behavior is ok for these cases. I played with

```python
import jax
import numpy as np
from jaxoplanet.experimental.starry import light_curves, Map

y = np.linspace(-2, 2, 100)

map = Map(y=None, period=1.2, u=None)

flux = jax.vmap(light_curves.map_light_curve, in_axes=(None, None, None, 0, None, None))(
    map, 0.1, 0.0, y, 2.0, 0.0
)
```

Let me know if you think about different tests.

src/jaxoplanet/experimental/starry/light_curves.py
lgrcia: Same here.

src/jaxoplanet/experimental/starry/light_curves.py
lgrcia: I agree that returning a `1+n_bodies` shape array (current) would play nice here

dfm: I'm inclined to move this conversation to a new issue (@lgrcia and I have identified another related problem to talk about), and get this PR merged since @dgegen has fixed the remaining bug and the results are consistent with what we had previously. I'll tag everyone over in the new issue!

PR #145, opened 2024-03-01
Adding @shashankdholakia to zenodo.json

dfm:
In https://github.com/exoplanet-dev/jaxoplanet/pull/87 we forgot to add @shashankdholakia to `.zenodo.json`. This fixes that.

@shashankdholakia: can you check that all the info that I added is right?


dfm:
Fixed. Thank you!

PR #146, opened 2024-03-01
Fix `test_compare_starry_system`

dgegen:
- Add `extract_starry_map_kwargs` to build default starry map if jaxoplanet's map is None
- Compare the corresponding light curves


dfm:
Thanks!! I don't know what's going on with GitHub actions today (the status page says that they're investigating issues), but I'll merge this as soon as we get CI back!

PR #149, opened 2024-03-01
Add `dgegen` to `zenodo.json`, fix syntax

dgegen:
Adding myself into `.zenodo.json`, as well as some missing commas to align the file with the json format.


dfm:
Thanks!

PR #151, opened 2024-03-01
feat: allow none args in map_light_curve

lgrcia:
Not essential but this PR allows to have `None` arguments for the occultor in `map_light_curve`. Useful to compute only the rotational light curve. It also solves an edge case in `float32`.


src/jaxoplanet/experimental/starry/light_curves.py
```diff
@@ -81,42 +79,55 @@ def light_curve_impl(time: Quantity) -> Array:
...
 def map_light_curve(
-    map: Optional[Map], r: Array, xo: Array, yo: Array, zo: Array, theta: Array
-) -> Array:
-    """Light curve of a map
+    map,
+    r: float = None,
+    xo: float = None,
+    yo: float = None,
+    zo: float = None,
+    theta: float = 0.0,
```
dfm, line 87: These should be typed as `Optional[Array]` rather than `float`.

dfm (changes requested):
Tiny change suggested, but otherwise looks great to me!

dfm (approved):

dfm:
The failure here looks like an unrelated numpyro / numpyro_ext compatibility issue because a new version of numpyro was released yesterday.

soichiro-hattori:
@lgrcia: I hope you don't mind but I resolved the conflict with the `main` branch so that this can be merged! I think we should merge if you think it's ready!

PR #156, opened 2024-03-03
Minor starry fixes

soichiro-hattori:
This PR fixes some minor bugs and docs issues in the starry module.

1. Update doc for `inc` argument in `Map` class to mention that it's initialized at 90 degrees (not 0) and that the angle is measured relative to the line of sight.
2. Add `self.normalize=normalize` in `Map` initialization so that it's string representation isn't `None`.
3. Fix doc mentions of `relative` in `Ylm` since we call it `normalize`.
4. In `Map` initialization, when in `normalize=True` block, cast `y[(0,0)]` to `float` as it's otherwise a JAX `ArrayImpl` that doesn't show the value in the string representation.

### DIFF
--- a/src/jaxoplanet/experimental/starry/maps.py
+++ b/src/jaxoplanet/experimental/starry/maps.py
@@ -70,6 +70,9 @@ class Map(eqx.Module):
     # Amplitude of the map, a quantity proportional to map luminosity.
     amplitude: Array
 
+    # Boolean to specify whether the Ylm coefficients should be normalized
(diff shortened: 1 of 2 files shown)


soichiro-hattori:
Now with changes that fix #157. 
@lgrcia I've removed `normalize` as an `Ylm` initialization argument and added a `.normalize()` method that returns a new `Ylm` instance with the normalized coefficients.

lgrcia (approved):

PR #159, opened 2024-03-04
Fixing type annotations for starry Map and removing default period

dfm:
It doesn't really make sense to have a default period of 10^15, when we could instead just handle the case of no rotation using a condition. I fixed that and a few tiny typing issues. Take a look, @lgrcia!


lgrcia (approved):

lgrcia:
You're right, looks good to me!

Issue #162, opened 2024-03-05
Renaming `SurfaceMapSystem` attributes


soichiro-hattori:
For a `SurfaceMapSystem` instance, the `central` object's surface map is called `surface_map` while the `bodies`' surface maps are called `surface_maps`. Could we rename them to `central_surface_map` and `bodies_surface_maps` for clarity? 

@dfm, @lgrcia: What do you think?

lgrcia:
I think it would work well!

dfm:
Sure - I'd be happy with that!

PR #165, opened 2024-03-06
Renaming SurfaceMapSystem surface map attributes

soichiro-hattori:
This PR implements the changes discussed in #162 and also updates some of the tests so that they work with the new attribute names.

@dfm, @lgrcia: I think it should be ready to merge but let me know if there's something I'm missing!


soichiro-hattori:
@dfm, @lgrcia: Should be ready to go but let me know if there's something I'm missing!

PR #167, opened 2024-03-11
Add tutorial introducing JAX

dgegen:
Motivated by the list of desired tutorials provided in [Notes from jaxoplanet docathon](https://github.com/exoplanet-dev/jaxoplanet/discussions/105), I wrote an introductory tutorial to JAX, trying to illustrate the most important features of JAX while keeping computationally intensive performance comparisons to a minimum, resulting in an execution time of ~5 seconds on my laptop.

Feedback is greatly appreciated. In particular with regards to
- Length: Do you think the tutorial is too short or too long?
- Content: Are there important features missing that you'd like to see covered?
- Placement: Is the location within the list of tutorials satisfactory, or should it be moved?

### DIFF
--- a/docs/tutorials/introduction-to-jax.ipynb
+++ b/docs/tutorials/introduction-to-jax.ipynb
@@ -36,6 +36,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "from typing import Union\n",
+    "\n",
     "import jax\n",
     "import jax.numpy as jnp\n",
     "import numpy as np\n",
@@ -88,7 +90,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "def sort_array(arr, library_name: str = \"\") -> None:\n",
+    "def sort_array(arr: Union[np.ndarray, jax.Array], library_name: str = \"\") -> None:\n",
     "    return_arr = arr.sort()\n",
     "    print(f\"{library_name:>5s}:  arr = {arr}, return_arr = {return_arr}\")\n",
     "\n",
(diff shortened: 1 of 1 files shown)


docs/tutorials/introduction-to-jax.ipynb
```diff
@@ -0,0 +1,646 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import jax\n",
+    "import jax.numpy as jnp\n",
+    "import numpy as np\n",
+    "\n",
+    "x = jnp.linspace(0, 5, 11)\n",
+    "x"
```
soichiro-hattori, line 44: `print(x)`

soichiro-hattori:
@dgegen: I think this is an awesome tutorial! I'm wondering if you think there's somewhere where you could include a snippet about how JAX handles out-of-bounds indexing? Or if not even just linking https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#out-of-bounds-indexing?

dgegen:
Thank you for your suggestion and kind words, @soichiro-hattori! I will revisit the tutorial soon to see where information about out-of-bounds indexing would best fit into the flow of information. Cheers!

docs/tutorials/introduction-to-jax.ipynb
```diff
@@ -0,0 +1,646 @@
...
+    "Let's look at a case NumPy's impure implementation leads to differing behavior between JAX and NumPy:\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def sort_array(arr, library_name: str = \"\") -> None:\n",
```
soichiro-hattori, line 91: For consistency, add data type for `arr`?

dgegen: This is indeed a somewhat awkward inconsistency that I could have easily avoided (`jax.typing.ArrayLike`). Thanks for bringing it to my attention :)

docs/tutorials/introduction-to-jax.ipynb
dgegen: The reason why I preferred to display `x` instead of printing it is that displaying also shows the type `float32`, which `print(x)` would not. Why would you favour `print(x)`?

dgegen:
@soichiro-hattori: I added the missing type hints as well as a small section on out-of-bounds indexing at the end of `An accelerated NumPy`.

Regarding the length of the tutorial:
I tried to present the most important functions of JAX in a short and structured way. But of course the length of the tutorial depends on the choice of the target group and the teaching objectives.
So by classifying the tutorial on the "slightly longer side" of the spectrum, do you mean that in comparison with the other tutorials or in reference to the word "brief" in its title? Or do you suggest covering less features? Or the same features, but more concisely? Or was it just an intuitive classification, perhaps rooted in a subconscious recognition of the tension between our desire for short tutorials and the fact that learning the most important JAX functions takes some time? :))

soichiro-hattori:
Hi @dgegen, I'm sorry for the late reply there's been a lot happening the past few weeks!

Thanks for adding the changes! 

>> Or was it just an intuitive classification, perhaps rooted in a subconscious recognition of the tension between our desire for short tutorials and the fact that learning the most important JAX functions takes some time? :))

Yes haha It was more of an intuitive classification! I don't see an issue with the length and I think it's great! 

I'm happy for this to be merged! @lgrcia, @catrionamurray?

dfm:
Thanks @dgegen for your contribution, and thanks @soichiro-hattori for the careful review!

PR #170, opened 2024-03-21
refactor: map to surface

lgrcia:
We recently renamed all "map" occurrences to "surface_map" in order to avoid using Python built-in `map` function. But we still have a `Map` object, and our new suffix creates some lengthy names like `central_surface_map` or `bodies_surface_maps`.

What about refactoring everything to "surface" only? Also turning the `Map` to a `Surface` object.

It's very minor but lightens the all syntax.


soichiro-hattori:
@lgrcia: That sounds good to me. 
Also, slightly pedantic but how do we feel about using `body_surfaces` instead of `bodies_surfaces`? I think it might be easier if we just always use `body_X` as the format (except the `bodies` iterable containing the multiple bodies in `SurfaceMapSystem`)? @dfm

soichiro-hattori:
@lgrcia: I don't have a strong opinion about this but do we need a visualization test as part of the required tests? I guess I'm wondering since it doesn't seem like we could check the figure anyways. But I don't have a good idea about it so I'm okay either way! 
We would need to add `matplotlib` to the `pyproject.toml` file's `test` dependencies though right?

Unrelated, but is there a reason why `import matplotlib` is inside the `show_map` function?

tests/experimental/starry/vis_test.py
```diff
@@ -0,0 +1,16 @@
+import numpy as np
+
+from jaxoplanet.experimental.starry.orbit import SurfaceBody
+from jaxoplanet.experimental.starry.surface import Surface
+from jaxoplanet.experimental.starry.visualization import show_map
+from jaxoplanet.experimental.starry.ylm import Ylm
+
+
+def test_show_map():
```
dfm, line 9: It looks like we'll need to install `matplotlib` for this test to work.

tests/experimental/starry/vis_test.py
```diff
@@ -0,0 +1,16 @@
...
+    y = Ylm.from_dense(np.hstack([1.0, np.random.rand(10) * 1e-1]))
+    surface = Surface(y=y, inc=0.9, obl=-0.3, period=1.2, u=[0.5, 0.5])
+    surface_body = SurfaceBody(period=1.0, surface=surface)
+
+    show_map(surface_body)
+    show_map(surface)
+    show_map(y)
```
dfm, line 16: This isn't really related to this PR, but it might be nice to check that this actually shows what we expect it to, rather than just checking that it displays?

src/jaxoplanet/experimental/starry/visualization.py
```diff
@@ -1,12 +1,12 @@
 import numpy as np
 
-from jaxoplanet.experimental.starry.maps import Map
+from jaxoplanet.experimental.starry.surface import Surface
 from jaxoplanet.experimental.starry.utils import graticule
 from jaxoplanet.experimental.starry.ylm import Ylm
 
 
 def show_map(
```
dfm, line 8: Should this be renamed to `show_surface`?

dfm (changes requested):
I'm totally on board with this renaming! Thanks @lgrcia!!

tests/experimental/starry/vis_test.py
lgrcia: I think we could skip it if `matplotlib` isn't installed, because that is tested in the `comparison` tests (`starry` requires `matplotlib`). Or is that a wrong practice to rely on other test environments?

src/jaxoplanet/experimental/starry/visualization.py
lgrcia: Absolutely! I will refactor

tests/experimental/starry/vis_test.py
dfm: Sure that would be fine! We should add a `pytest.importorskip` in that case then.

tests/experimental/starry/vis_test.py
lgrcia: Any idea of how to do that? Saving an array of the expected result seems overkill right?

dfm: You could take a look at the tests in corner.py which follows the matplotlib approach: https://github.com/dfm/corner.py/tree/main/tests

But that might be overkill for our case. Your call!

lgrcia:
@soichiro-hattori and @dfm, please read the comments above but my plan was to keep `matplotlib` out of the dependencies and use `pytets.importorskip('matplotlib')` to fix the tests

dfm:
I'm fine with that plan, @lgrcia!

dfm (approved):

PR #179, opened 2024-05-07
Improve interface of `TransitOrbit`

dgegen:
This pull request aims make the interface of `TransitOrbit` less confusing and more user-friendly.

## Changes Made
- Extended the docstring.
- Renamed the `radius` parameter to `ror` (Ratio of planet radii to star radius). Reasoning: Using the term `radius` for what is in practice the unitless ratio of planetary radii to stellar radii is a rather confusing nomenclature - even for developers, as the comment `#TODO: Is it actually the radius ratio?` in the [Transit Fitting tutorial](https://jax.exoplanet.codes/en/latest/tutorials/transit/) illustrates to the user seeking to unravel the usage of this class. As far as I can tell, this attribute was originally named `radius` to ensure formal consistency with the Keplerian orbit class (although it should be noted that the two differ in that the `radius` attribute of that class has units of length). To remove this confusion, `radius` has been replaced by the attribute `ror`, and a `radius` property returning `ror` has been added to ensure comparability with other functions, such as `limb_dark_light_curve`, in the background.

We could consider temporarily keeping the `radius` as an additional attribute to ensure backward comparability, but in the interest of clear unambiguous nomenclature, I think it is best not to do so.

As always, feedback is greatly appreciated :))


soichiro-hattori:
This looks good to me @dgegen! Thank you for doing this!

I'm happy to merge! @lgrcia?

dgegen:
All good :))

I agree that consistency between interfaces is nice. In this case, however, I am not sure if it is not more preferable to have clear and self-explanatory variable names. I think `ror` or `radius_ratio` would be more explicit, highlight the conceptual difference between the usage in this class and in the keplerian orbit class, where the radius field has a units field that is part of the implementation, and eliminate the need to consult the documentation.

On the other hand, if we were looking for maximum consistency, we would probably allow the user to change the value of central_radius, and make both radius and central_radius have units. Then again, this adds complexity that is not really helping anyone.

soichiro-hattori:
I think I agree with both of you here! Consistency in the interface is nice but I also think it can be a bit confusing if `radius` means slightly different things between the two `orbit` objects. 

I don't know if this makes is convoluted but how do you two feel about allowing for either `radius` and `ror` to be used (we'd check to ensure only one of them is passed)? We could default to using `ror` in the docs and in general but also maintain some interface consistency by allowing the user to also use `radius`? @dgegen, @lgrcia

lgrcia:
Sorry for the delay here. Back on jaxoplanet! Let's name it `radius_ratio` for now and we'll discuss more if needed when preparing a new release.

dgegen:
No bother at all! I finally got around to renaming all relevant instances of `radius` to `radius_ratio`, so in principle this pull request can be merged/rebased now :))

PR #197, opened 2024-08-09
feat: order as kwarg and some ylm utilities + tests

lgrcia:


dfm (approved):
LGTM, although `order` is a bit overloaded throughout the code. @soichiro-hattori and I have flip flopped on this, and I think it's fine to just use `order` here, but we should make sure it gets documented!

PR #203, opened 2024-08-14
feat: diagonal sparse Pijk + discussion on merging limb-dark and starry

lgrcia:
This PR is an attempt to understand what can be done to **merge the limb-darkening light curve implementation with starry**.

## Motivation
The idea is that, if only limb-darkening is present on a map defined with starry, JAX should figure out which part of the different matrices are necessary to compute the starry light-curve, and that these should be pretty similar in number to the ones involved in `core.limb_dark.light_curve`. Hence performances of the starry light curves should be similar to the limb-dark one, i.e. `experimental.starry.light_curves.surface_light_curve` with
```python
surface = Surface(y=None, u=u)
```

## Description of the current modifs
In the `Ylm` and `Pijk` interfaces, we defined a `diagonal` attribute which indicates if the *non-radial* coefficients of the spherical harmonics basis, i.e. not contributing to limb-darkening, are all zeros. The first commit in this PR fixes that for the `Pijk` basis, and allows to pass a `diagonal` kwarg to the `Pijk.to_sparse` method, so that the computation

```python
# see end of jaxoplanet/experimental/starry/light_curves.py

p_y.tosparse(diagonal=only_u) @ design_matrix_p
```
 takes into account the sparsity of `p_y`.
 
## Results
For now this is not working. The starry light curve is ~ 4 times slower than the limb-dark one.

**starry**
```python
import jax
from jaxoplanet.experimental.starry.surface import Surface
from jaxoplanet.experimental.starry.light_curves import surface_light_curve
import numpy as np

r = 0.1
u = (0.1, 0.2)
b = np.linspace(0, 1 + r, 1000)
order = 20

surface = Surface(u=u)
function = jax.jit(jax.vmap(lambda b: surface_light_curve(surface, r, z=10.0, y=b, order=order)))
jax.block_until_ready(function(b))
%timeit function(jax.block_until_ready(function(b)))
```
```
439 μs ± 2.82 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
**limb-dark**
```python
from jaxoplanet.core.limb_dark import light_curve

function = jax.jit(jax.vmap(lambda b: light_curve(u, b, r, order=order)))
jax.block_until_ready(function(b))
%timeit function(jax.block_until_ready(function(b)))
```
```
115 μs ± 2.04 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

[...]

PR #204, opened 2024-08-17
Optimizing starry impl

dfm:
These optimizations start to narrow the gap between the limb_dark and starry solution vector computational performance. The main differences:

1. Using the closed form solution for s0 instead of the numerical one.
2. In the previous implementation a _lot_ of zeros were being summed in P integral for the diagonal case. The key is to just never compute the off diagonal `arg`s, and only update the indices accordingly.

Still to do:

1. We should be able to use the closed form s2 as well, but it wasn't clear to me where it would go.
2. Figure out the shapes. I don't totally understand the logic used for determining the "diagonal" indices, but why do we get a 4 element solution vector for `order=2`, instead of 3 elements like in the limb_dark case?

### DIFF
--- a/src/jaxoplanet/experimental/starry/solution.py
+++ b/src/jaxoplanet/experimental/starry/solution.py
@@ -21,21 +21,67 @@ def solution_vector(
     def impl(b: Array, r: Array) -> Array:
         b = jnp.abs(b)
         r = jnp.abs(r)
-        kappa0, kappa1 = kappas(b, r)
-        P = p_integral(order, l_max, b, r, kappa0, diagonal=diagonal)
-        Q = q_integral(l_max, 0.5 * jnp.pi - kappa1, diagonal=diagonal)
-        return Q - P
+        area, kappa0, kappa1 = kappas(b, r)
+
+        no_occ = jnp.greater_equal(b, 1 + r)
+        full_occ = jnp.less_equal(1 + b, r)
+        cond = jnp.logical_or(no_occ, full_occ)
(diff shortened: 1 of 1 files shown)


lgrcia:
Thanks so much for having a look at that @dfm! 

Your two questions are linked. Agol basis is 

$$(1\\;z\\;z^2\\;z^3\\;\dots\\;z^n)^T\text{,}$$

and Luger's basis is 

$$(1\\;x\\;z\\;y\\;x^2\\;xz\\;xy\\;yz\\;y^2\dots)^T\text{.}$$

Luger's basis does not contain powers of $z$ greater than 1. So to convert vectors from Agol's basis to Luger's basis we have to use the fact that, on the unit sphere, $z^2 = 1 - x^2 - y^2$. This explains the difference in shapes between the two basis vectors and can be used to find the transformation from one basis to the other, so that the closed form s2 can be reused. I have this transformation written somewhere and can wrap that up next week!

soichiro-hattori:
Hey @dfm and @lgrcia,

I've been trying to get a better understanding of some of the proposed changes in this PR and #203, and how it relates to the failed tests.

1. Am I correct that the change in the `for` loop lower limit from 0 to 1 in both the `p_integral` and `q_integral` functions come from using the closed-form solution for the `s0` component?
2. The current code fails because of the shape mismatch for the `sT @ A2` operation in the `surface_light_curve` function. My guess is that it's due to how `sT` is now calculated using Agol's basis while `A2` is still in the Luger basis? @lgrcia: Do the newer commits in #203 fix this issue with another transformation?

Issue #207, opened 2024-08-31
Moving new release features to `main`


lgrcia:
A lot of features in the `core_paper` branch (which is the version used in the paper) have to be incrementally integrated in the `main` branch. This issue keeps track of that.

- [x] multi-precision core using `mpmath` (#210)
- [ ] ~~making a CI pipeline producing an artifact to monitor the numerical precision of the `main` implementation (and speed)~~
- [ ] ~~using the multi-precision $A_1$ matrix in starry light curve~~
- [x] using the symmetry of the solution vector integrand for `n>2` in order to speed up numerical integration (at least 30% speedup) (#217)
- [x] rotation matrices using `s2fft` (for now only a single edge-case grad test remains to be solved) (#212)
- [x] order of numerical integration as keyword argument in starry light-curve (#197)
- [x] Fixing starry light curves amplitudes of multiple bodies (#195)

lgrcia:
Update: We are trying to merge #212 and #217 but encounter issues with tests on macOs (only). We think it's related to the use of s2fft I'm currently trying to pin down the issue.

lgrcia:
All done in #227

PR #210, opened 2024-09-04
feat: add a multi-precision module for testing

lgrcia:
This adds a multi-precision module to compute all starry quantities using the mpmath package with arbitrary precision.

### DIFF
--- a/tests/precision/test_flux.py
+++ b/tests/precision/test_flux.py
@@ -3,7 +3,7 @@ import numpy as np
 import pytest
 
 from jaxoplanet.experimental.starry import light_curves
-from jaxoplanet.experimental.starry.multiprecision import flux as mp_flux, mp, utils
+from jaxoplanet.experimental.starry.multiprecision import flux as mp_flux, mp
 from jaxoplanet.experimental.starry.surface import Surface
 from jaxoplanet.experimental.starry.ylm import Ylm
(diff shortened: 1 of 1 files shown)


soichiro-hattori:
@lgrcia: I'm happy for this to be merged! 
Will these module need to be documented or is it only going to be used for testing?

lgrcia:
I think mainly testing. I also had in mind a CI (maybe not on GitHub) to monitor the precision after each PR. We can talk about it in #207.

lgrcia:
Before I merge, @dfm, do you think this is a meaningful thing to push to main? It's kind of a big addition only for testing so I'd love to have your feedback.

tests/precision/test_flux.py
```diff
@@ -0,0 +1,138 @@
...
+        ha="left",
+        va="top",
+        alpha=0.75,
+    )
+    plt.tight_layout()
+    plt.savefig("flux_precision.png")
+
+
+if __name__ == "__main__":
+    plot_flux_precision()
```
dfm, line 138: I'm not 100% sure I'd put these plotting functions in `tests`. Maybe `examples` or `scripts` or something? Or maybe at least split all the plotting tests out into a separate script?

dfm (approved):
LGTM! Tiny comment inline, but this makes sense to me!

PR #211, opened 2024-09-09
Fixed relative angles unpacking bug

shishirdholakia:
Seems like there was an unpacking bug in relative angles


lgrcia:
@shishirdholakia, thanks for this fix! Would that be possible to add a very simple unit-test for this function?

soichiro-hattori:
@lgrcia: Should we go ahead and merge this PR?

lgrcia:
Sure, let's do that! Thanks for having a look. 

I will try and merge taking out the starry experimental after that.

PR #225, opened 2024-10-09
feat: `None` values for angles

lgrcia:
This PR enables `None` values for the inclination and obliquity of a surface, enabling loads of optimizations when preforming its rotation.


lgrcia:
By mistake, this contains the code from #212  and #217 ...

Issue #229, opened 2024-10-26
Jax version update, units issue


ben-cassese:
Hi! It looks like jax versions >=0.4.34 introduced a change that's not playing well with units in jaxoplanet. Here's an example:

```python
import jax
jax.config.update("jax_enable_x64", True)
import jax.numpy as jnp

import jaxoplanet
from jaxoplanet.orbits.keplerian import Central, Body, System
from jaxoplanet.units import unit_registry as ureg
from jaxoplanet import units
from jaxoplanet.light_curves import limb_dark_light_curve

print(f"jax version: {jax.__version__}")
print(f"jaxoplanet version: {jaxoplanet.__version__}")

t = jnp.linspace(-1, 1, 1728) * ureg.day  # 100s cadence for 48 hours
poly_limbdark_order = 2
key = jax.random.PRNGKey(0)

key, *subkeys = jax.random.split(key, num=9)

star_mass = jax.random.uniform(subkeys[0], minval=0.1, maxval=1.5) * ureg.M_sun
semimajor_axis = jax.random.uniform(subkeys[1], minval=0.005, maxval=5.0) * ureg.au
impact_param = jax.random.uniform(subkeys[2], minval=0.0, maxval=1.0)
planet_rad = jax.random.uniform(subkeys[3], minval=0.001, maxval=0.25) * ureg.R_sun
eccentricity = jax.random.uniform(subkeys[4], minval=0.0, maxval=0.9)
omega = jax.random.uniform(subkeys[5], minval=0.0, maxval=2 * jnp.pi)
u = jax.random.uniform(subkeys[6], shape=(poly_limbdark_order,))
star_radius = jax.random.uniform(subkeys[7], minval=0.1, maxval=1.5) * ureg.R_sun
Omega = jnp.pi


star = Central(radius=star_radius, mass=star_mass)
planet = (
    System(star)
    .add_body(
        time_transit=0.0,
        semimajor=semimajor_axis,
        impact_param=impact_param,
        radius=planet_rad,
        eccentricity=eccentricity,
        omega_peri=omega * ureg.rad,
        asc_node=Omega * ureg.rad,
        mass=0.0,
    )
    .bodies[0]
)
```

This produces the output error:

```python
jax version: 0.4.35
jaxoplanet version: 0.0.2

ValueError                                Traceback (most recent call last)
Cell In[1], line 34
     28 Omega = jnp.pi
     31 star = Central(radius=star_radius, mass=star_mass)
[...]

PR #230, opened 2024-10-28
Hotfix PR229

soichiro-hattori:
This should solve the bug that's mentioned in #229.
The fix is to just create an `if else` branch in the `converter(value)` function to return `None` if `value=None` and otherwise to use the `tree.map` functionality to apply units.
   
@dfm: Could you let me know if these changes are the right way to handle this bug?


soichiro-hattori:
`tree.map` is changed back to `tree_util.tree_map` for backwards compatibility.

PR #251, opened 2025-02-03
Enabling Transit Timing Variations (TTV) Fitting with Jaxoplanet

TylerFair:
This is essentially a carbon copy of exoplanet's TTVOrbit integrated into the jaxoplanet architecture. I will note I am definitely not an expert at jax, and therefore this model will be slightly unoptimized, but from my benchmark tests, the model performs reasonably well at 10,000 points with ~50 transits between two planets with a convergence time of ~10 mins. One small thing is that for a reason unknown to me the model plummets to target_accept values of 0.00 if dense_mass is set to True, so setting it to false seems to be a must for high-parameter solutions, which I assume is why it is defaulted False in numpyro.infer. 

To explain the module briefly: This implementation provides a framework for handling Transit Timing Variations (TTVs) by extending the base TransitOrbit class with time-warping capabilities. This is in a sense identical to TTVOrbit from Exoplanet which builds off of KeplerianOrbit. The core functionality remains the same, enabling users to either directly input TTV offsets or provide observed transit times, from which the code automatically computes the linear ephemeris and ttvs through linear regression. Using a binning-based lookup system, it maps arbitrary observation times to their corresponding (possibly shifted) transit events, handling both single and multi-planet configurations. The code maintains full compatibility with TransitOrbit while adding TTV-specific features like period/t0 inference, transit time tracking, and automatic unit handling (hopefully). 

This should presumably go into an experimental folder if deemed clear to try until completely cleared but pushing to typical location anyway. Please feel free to make any changes you seem necessary or reject if you believe it should be built differently. 

I have tested this on some previous case studies of mine and it appears to work well for known TESS systems thusfar.


TylerFair:
Just to note, made last improvements to ttv script. It now correctly works for all inputs of transit_times, transit_times+log_delta_period and ttvs+period. I've also removed most of the for loops and enabled more jitting functionality than previously, also improved the units logic a little and removed some redundancies. This should overall be a lot more polished version for inspection. 

I've also updated the tutorial notebook a little bit for some minor additional comments and model changes. 

These are the last PR changes I'll make on this but hopefully it may prove helpful in future!

lgrcia:
Thanks so much @TylerFair and sorry for the delay on this. I'll have a look by the end of the week!

docs/tutorials/ttv.ipynb
```diff
@@ -0,0 +1,513 @@
...
+    ")\n",
+    "\n",
+    "sampler.run(jax.random.PRNGKey(1), time, yerr, y=y)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let's print the summary of our sampling, [as explained here](./transit.ipynb/#checking-our-posterior-samples), focusing on $\\hat{R}$ and the number of divergences encountered."
```
lgrcia, line 379: The reference `[as explained here](./transit.ipynb/#checking-our-posterior-samples)` cannot be resolved and lead to the doc build to fail.

Could you remove this part all together? Just leaving:
> Let's print the summary of our sampling, focusing on $\\hat{R}$ and the number of divergences encountered.

Sorry this was one of my addition in your fork PR.

lgrcia (commented):
This looks great. Once the tests pass and the tutorial is fixed I am happy to merge!

docs/tutorials/ttv.ipynb
TylerFair: Resolved this line by removing the mentioned problematic reference.

TylerFair:
Let me know if that change fixed it! Thanks for the checks and revision of both codes!

PR #257, opened 2025-02-18
feat: radial velocity of spherical harmonics maps

lgrcia:
This adds functions to compute the RV of a map described by SH.

Here is what's left to do:
- [x] make `surface_radial_velocity` function
- [x] test `surface_radial_velocity` against starry
- [x] make radial velocity function of system
- [ ] test value and gradient
- [x] make plotting function like starry


lgrcia:
I added a tutorial similar to https://starry.readthedocs.io/en/latest/notebooks/RossiterMcLaughlin/. I also added a "conventions" page that I'll populate in another PR. It contains some useful plots of the conventions we use when dealing with spherical harmonics and some coordinate systems.

PR #262, opened 2025-03-19
Remove units

lgrcia:


soichiro-hattori:
This PR deprecates supporting units in `jaxoplanet`. This also allows us to remove the dependencies on the `jpu` and `pint` packages.

Additional changes:
- `TransitOrbit.relative_position()` returns `jnp.stack([x, y, z])` so that it has the correct shape for when used directly and also when called from `TTVOrbit.relative_position()`. 
- Alongside the above change, we removed the `@vectorize` decorator for the `light_curve_impl` closure/inner function. Instead, we simply `vmap` the closure function (i.e., `jax.vmap(light_curve_impl)`)
- We set `regularize_mass_matrix=True`, `dense_mass=True` when we perform MCMC sampling in the tutorials since we seem to now be efficient enough even with those enabled.

Issue #264, opened 2025-03-31
power2 limb-darkening?


saigrain:
Has anyone tried to implement a power2 limb-darkening law (cf Maxted 2018, Maxted & Gill 2018) in jaxoplanet (or even exoplanet)? Or is there any plan to do so? My understanding is that it gives a much better match to observed limb-darkening profiles near the limb than polynomial laws.

lgrcia:
Hi @saigrain, thanks for asking. There is no plan to implement it although it could be very interesting to have it!

I see the implementation is fairly easy (code provided in https://ui.adsabs.harvard.edu/abs/2022A%26A...664A.128C/abstract). Let me know if you are working on an implementation on your side or I am also happy to implement it and open a pull request!

soichiro-hattori:
@lgrcia: I may be wrong here but since this is a non-polynomial limb-darkening parameterization, would you implement this by generating it in the power2 parameter space but then fitting it with a high-order polynomial so that it works within the arbitrary polynomial limb-darkening framework we have?

lgrcia:
@soichiro-hattori, you are totally right, I had the same thought. But we don't have to implement it using starry at all. Check out the appendix in [this paper](https://arxiv.org/abs/1812.01606) and you'll see that the code is fairly simple. The light curve has a closed form and can be implemented with `jax.numpy` functions. 

I pushed some working code to the `power2` branch but I'd be very happy for @saigrain to open a PR if she wants to give the implementation a try!

EDIT: sorry wrong paper, corrected the url!

saigrain:
Amazing, that was quick! I hadn't started my own implementation, so I will try testing the branch @lgrcia started.

PR #265, opened 2025-04-01
feat: add power2 light curve

lgrcia:
As suggested by #264; adds a light curve model with the power-2 limb-darkening law.

For now comparison with pytransit fails (see below). I haven't tried to debug at all but some help would be great!
```python
import jax

import numpy as np
from jaxoplanet.orbits.keplerian import System
from jaxoplanet.light_curves import limb_dark_light_curve, power_two_light_curve
from pytransit import RRModel

time = np.linspace(-0.1, 0.1, 1000)

tm = RRModel("power-2")
tm.set_data(time=time)

radius = 0.2
system = System().add_body(period=1.0, radius=radius)

coeff = (0.4, 1.0)
expected = tm.evaluate(
    k=radius,
    ldc=coeff,
    t0=0.0,
    p=system.bodies[0].period.magnitude,
    a=system.bodies[0].semimajor.magnitude,
    i=np.pi / 2,
)

calc = power_two_light_curve(system, *coeff, eps=1e-15)(time).T[0] + 1.0

plt.subplot(121)
plt.plot(time, expected, label="pytransit")
plt.plot(time, calc, label="jaxoplanet")
plt.legend()
plt.xlabel("time")
plt.ylabel("flux")

plt.subplot(122)
plt.plot(time, expected - calc)
plt.xlabel("time")
plt.ylabel("error")
plt.tight_layout()
```
Screenshot 2025-04-01 at 12 38 15 PM

Suggested steps to reproduce the environment ([uv](https://github.com/astral-sh/uv) required):
```
git clone -b power2 https://github.com/exoplanet-dev/jaxoplanet.git
cd jaxoplanet
uv sync --extras dev
uv pip install pytransit # pytransit for comparison
source .venv/bin/activate
```

@saigrain, do you know what precision is expected from `pytransit`?

### DIFF
--- a/src/jaxoplanet/core/power_two.py
+++ b/src/jaxoplanet/core/power_two.py
@@ -1,0 +1,70 @@
+from functools import partial
+
+import jax
+import jax.numpy as jnp
+
+
+@partial(jax.jit, static_argnames=("eps",))
+def light_curve(c, alpha, b, r, eps=1e-12):
+    I_0 = (alpha + 2) / (jnp.pi * (alpha - c * alpha + 2))
+    g = 0.5 * alpha
+    b = jnp.abs(b)
+    r = jnp.abs(r)
+
+    def q1(z, p, c, alpha):
+        zt = jnp.clip(abs(z), 0, 1 - p)
+        s = 1 - zt**2
+        c0 = 1 - c + c * s**g
+        c2 = 0.5 * alpha * c * s ** (g - 2) * ((alpha - 1) * zt**2 - 1)
+        return 1 - I_0 * jnp.pi * p**2 * (
+            c0 + 0.25 * p**2 * c2 - 0.125 * alpha * c * p**2 * s ** (g - 1)
+        )
+
+    def q2(z, p, c, alpha):
+        zt = jnp.clip(abs(z), 1 - p, 1 + p)
+        d = jnp.clip((zt**2 - p**2 + 1) / (2 * zt), 0, 1)
+        ra = 0.5 * (zt - p + d)
+        rb = 0.5 * (1 + d)
+        sa = jnp.clip(1 - ra**2, eps, 1)
+        sb = jnp.clip(1 - rb**2, eps, 1)
+        q = jnp.clip((zt - d) / p, -1, 1)
+        w2 = p**2 - (d - zt) ** 2
+        w = jnp.sqrt(jnp.clip(w2, eps, 1))
+        b0 = 1 - c + c * sa**g
+        b1 = -alpha * c * ra * sa ** (g - 1)
+        b2 = 0.5 * alpha * c * sa ** (g - 2) * ((alpha - 1) * ra**2 - 1)
+        a0 = b0 + b1 * (zt - ra) + b2 * (zt - ra) ** 2
+        a1 = b1 + 2 * b2 * (zt - ra)
+        aq = jnp.arccos(q)
+        J1 = (
+            a0 * (d - zt)
+            - (2 / 3) * a1 * w2
+            + 0.25 * b2 * (d - zt) * (2 * (d - zt) ** 2 - p**2)
+        ) * w + (a0 * p**2 + 0.25 * b2 * p**4) * aq
+        J2 = (
+            alpha
+            * c
+            * sa ** (g - 1)
+            * p**4
+            * (
+                0.125 * aq
+                + (1 / 12) * q * (q**2 - 2.5) * jnp.sqrt(jnp.clip(1 - q**2, 0, 1))
+            )
+        )
+        d0 = 1 - c + c * sb**g
+        d1 = -alpha * c * rb * sb ** (g - 1)
+        K1 = (d0 - rb * d1) * jnp.arccos(d) + (
+            (rb * d + (2 / 3) * (1 - d**2)) * d1 - d * d0
+        ) * jnp.sqrt(jnp.clip(1 - d**2, 0, 1))
+
+        K2 = (1 / 3) * c * alpha * sb ** (g + 0.5) * (1 - d)
+        return 1 - I_0 * (J1 - J2 + K1 - K2)
+
+    return (
+        jnp.select(
+            [b <= (1 - r), abs(b - 1) < r],
+            [q1(b, r, c, alpha), q2(b, r, c, alpha)],
+            default=1,
+        )
+        - 1.0
+    )
--- a/src/jaxoplanet/light_curves/__init__.py
+++ b/src/jaxoplanet/light_curves/__init__.py
@@ -1,4 +1,4 @@
-__all__ = ["transforms", "limb_dark_light_curve"]
+__all__ = ["transforms", "limb_dark_light_curve", "power_two_light_curve"]
 
 from jaxoplanet.light_curves import transforms as transforms
 from jaxoplanet.light_curves.limb_dark import light_curve as limb_dark_light_curve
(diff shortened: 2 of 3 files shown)


lgrcia:
An update on this; The expression for the disk integral in https://arxiv.org/pdf/1812.01606 is obtained through a Taylor expansion, so the precision of the light curve will be pretty limited. From the [pycheops](https://github.com/pmaxted/pycheops/blob/a7bbf9526fe0f558bff62f6aaef70a7f965e50cb/pycheops/models.py#L57) code:
> The approximation is accurate to better than 100ppm for radius ratio k < 0.1.
> **N.B.** qpower2 is untested/inaccurate for values of k > 0.2

Below is a more robust test against numerical integration. jaxoplanet is under the precision bar but still less precise than pytransit for some reason.

<details>

<summary>source code for the plot</summary>

```python
from scipy.integrate import dblquad
import jax

import numpy as np
from jaxoplanet.orbits.keplerian import System
from jaxoplanet.light_curves import limb_dark_light_curve, power_two_light_curve
from pytransit import RRModel

time = np.linspace(0.0, 0.05, 40)

tm = RRModel("power-2")
tm.set_data(time=time)

radius = 0.1
system = System().add_body(period=1.0, radius=radius)

coeff = (0.4, 1.0)
expected = tm.evaluate(
    k=radius,
    ldc=coeff,
    t0=0.0,
    p=system.bodies[0].period.magnitude,
    a=system.bodies[0].semimajor.magnitude,
    i=np.pi / 2,
)

calc = power_two_light_curve(system, *coeff, eps=1e-6)(time).T[0] + 1.0


def _numerical_flux(b, r, epsabs=1e-15, epsrel=1e-15, I=None):
    """Compute the flux by numerical integration of the surface integral."""
    # I'm only coding up a specific case here
    assert (b >= 0) and (r <= 1), "Invalid range."

    # Total flux
    total, _ = dblquad(
        I,
        -1,
        1,
        lambda x: -np.sqrt(1 - x**2),
        lambda x: np.sqrt(1 - x**2),
        epsabs=epsabs,
        epsrel=epsrel,
    )
    if b >= 1 + r:
        return total

    # Get points of intersection
    if b > 1 - r:
        yi = (1.0 + b**2 - r**2) / (2.0 * b)
        xi = (1.0 / (2.0 * b)) * np.sqrt(4 * b**2 - (1 + b**2 - r**2) ** 2)
    else:
        yi = np.inf
        xi = r

    # Lower integration limit
    def y1(x):
        if yi <= b:
            # Lower occultor boundary
            return b - np.sqrt(r**2 - x**2)
        elif b <= 1 - r:
            # Lower occultor boundary
            return b - np.sqrt(r**2 - x**2)
        else:
            # Tricky: we need to do this in two parts
            return b - np.sqrt(r**2 - x**2)

    # Upper integration limit
    def y2(x):
        if yi <= b:
            # Upper occulted boundary
            return np.sqrt(1 - x**2)
        elif b <= 1 - r:
            # Upper occultor boundary
            return b + np.sqrt(r**2 - x**2)
        else:
            # Tricky: we need to do this in two parts
            return np.sqrt(1 - x**2)

    # Compute the total flux
    flux, _ = dblquad(I, -xi, xi, y1, y2, epsabs=epsabs, epsrel=epsrel)

    # Do we need to solve additional integrals?
    if not (yi <= b) and not (b <= 1 - r):

        def y1(x):
            return b - np.sqrt(r**2 - x**2)

        def y2(x):
            return b + np.sqrt(r**2 - x**2)

        additional_flux1, _ = dblquad(I, -r, -xi, y1, y2, epsabs=epsabs, epsrel=epsrel)
        additional_flux2, _ = dblquad(I, xi, r, y1, y2, epsabs=epsabs, epsrel=epsrel)

        flux += additional_flux1 + additional_flux2

    return total - flux


def numerical_flux(alpha, c, b, r, epsabs=1e-8, epsrel=1e-8):
    def I(y, x):
        mu = np.sqrt(1 - x**2 - y**2)
        return 1 - alpha * (1 - mu**c)

    nf = partial(_numerical_flux, epsabs=epsabs, epsrel=epsrel, I=I)
    I0 = nf(10, 0.1)
    return np.vectorize(nf)(b, r) / I0 - 1


import jpu.numpy as jnpu
from jaxoplanet.units import unit_registry as ureg
from functools import partial


def numerical_light_curve(orbit, alpha, c, time):
    r_star = orbit.central_radius
    x, y, z = orbit.relative_position(time)
    b = jnpu.sqrt(x**2 + y**2) / r_star
    assert b.units == ureg.dimensionless
    r = orbit.radius / r_star
    assert r.units == ureg.dimensionless
    lc_func = partial(numerical_flux, alpha, c)
    lc = lc_func(b.magnitude, r.magnitude)
    return lc[0] + 1.0


truth = numerical_light_curve(system, *coeff, time)

plt.subplot(121)
plt.plot(time, expected, label="pytransit")
plt.plot(time, calc, label="jaxoplanet")
plt.plot(time, truth, label="numerical")
plt.title(f"radius_ratio={radius:.2f}")
plt.legend()
plt.xlabel("time")
plt.ylabel("flux")

plt.subplot(122)
plt.plot(time, truth - calc, label="jaxoplanet")
plt.plot(time, truth - expected, label="pytransit")
plt.axhline(100e-6, ls="--", c="k")
plt.xlabel("time")
plt.ylabel("error")
plt.tight_layout()
```

</details>
Screenshot 2025-04-01 at 2 22 59 PM

@soichiro-hattori, we may want to explore the approach of fitting the profile to the polynomial basis as you suggested. This way would could even provide a function to do so for any non-linear limb-darkening (if that's not already available in the exoplanet-codes family).

saigrain:
Thanks for this @lgrcia !

I think you swapped the labels of the two curves, otherwise it would imply the PyTransit version is less precise than the jaxoplanet one.

Just to check I understand correctly, the numerical calculation in your test is doing the disk integral calculation for a power2 LD profile, but without the Taylor approximation, right? A 30ppm deviation near the limb could be problematic for some applications, I guess, e.g. large-planet TTVs in PLATO data. Right now I'm working on small planets, I suspect it won't be such a problem.

About approximating the power2 law with polynomials: the power2 law is popular because it gives a better approximation to the intensity profiles produced by 3D atmosphere models with fewer parameters than a polynomial. I attach plots provided by my collaborator Lionel Bigot who generated (for a Sun-like case) an intensity profile and fit it with a polynomial law on one hand and a power2 law on the other. The differences are small but the polynomial version requires at least order 4 for a decent fit. 

polynomial_LD
power2_LD

So if you were to approximate the power2 law with polynomials, I guess you'd have to go to a relatively high order, and wouldn't that defeat the point? Maybe I misunderstand how it would work

soichiro-hattori:
@lgrcia: We could definitely incorporate that in. I already have the code necessary to do this given the recent work I've been doing with limb darkening. I suppose we should chat more about where that code would reside within this package. 

@lgrcia, @saigrain: I also wanted to highlight that this higher-order polynomial fit to non-polynomial limb-darkening laws has been done in related packages such as my colleague @ben-cassese's [squishyplanet](https://github.com/ben-cassese/squishyplanet) (specifically covered [here](https://squishyplanet.readthedocs.io/en/latest/tutorials/limb_darkening_choices.html) in the docs).

saigrain:
@soichiro-hattori aha - having checked the squishyplanet docs, I see that if you're willing to go up to quite high polynomial orders the problem disappears - that might be the way to go then

Issue #256, opened 2025-02-13
Issue pip installing jaxoplanet


shishirdholakia:
The version of jax that is pinned in pyproject.toml is "jax<=0.4.31", but jaxlib does not appear to be pinned. As a result, when I try to pip install jaxoplanet and import it, I get:
```python
RuntimeError: jaxlib version 0.4.38 is newer than and incompatible with jax version 0.4.29. Please update your jax and/or jaxlib packages.```

lgrcia:
Thanks for spotting that @shishirdholakia! I'll make sure to pin both versions in the next PR and release.

dfm:
Remind me why we need the 0.4.31 pin? Can we try updating that again?

soichiro-hattori:
I'll try to get the pin removed, but in the mean time at least with `uv`, if one just runs 
`uv add jaxoplanet` then it'll likely fetch the `jaxoplanet==0.0.2` version since our installation instructions recommend installing `jax` first (and will likely give you `jax>0.5.2`).

If you want the `jaxoplanet==0.0.4` version (which I think we need), then the uv install command should be
`uv add "jaxoplanet==0.0.4"` (which will currently install `jax=0.4.31` which is a bit behind).

soichiro-hattori:
This issue should be resolved with the recent merge (#266).

PR #266, opened 2025-04-01
Remove jax pin

soichiro-hattori:
Related to #256.
Testing to see whether we still need the pin. 
It'd be nice to remove it since it creates some headaches when trying to install on new environments that have newer jax version.


soichiro-hattori:
The main changes in this PR:

1. Remove `jax<=0.4.31` pin in `pyproject.toml` since this causes installation issues (see #256).
2. Change `jpu` dependency in `pyproject.toml` to directly point to the [`jpu` github repo ](https://github.com/dfm/jpu) as the `PyPI` published version does not have a necessary commit included (see https://github.com/dfm/jpu/issues/25). This change should be reverted once the new `jpu` version is published.
3. The `test_EB()` test in [`tests/starry/light_curve_test.py`](https://github.com/exoplanet-dev/jaxoplanet/blob/main/tests/starry/light_curve_test.py) is marked as an `xfail` (expected failure) as we need to correct a bug (see #268). 

Note: The `numpy<1.22` pin for the `comparison` tests seems to be required to get a valid `PyMC3` version installed.

PR #319, opened 2026-05-01
Fix Rsun to arcsec conversion in Keplerian orbit with parallax 

vandalt:
I was trying to reproduce the [exoplanet astrometric fitting example](https://gallery.exoplanet.codes/tutorials/astrometric/) with `jaxoplanet` and encountered an issue with the separation being too large. It seems the issue was when converting r0 from Rsun (the units of `self.semimajor`) to arcsec, `jaxoplanet` was multipling by `constants.au` (the number of Rsun in an au) instead of dividing by it. This only happens when `parallax` is not `None`.

This PR adds a test for it and fixes the issue. I was wondering if it would be worth also testing this through a comparison with a well-tested astrometry package like orbitize! (this would require adding orbitize! as a test dependency).

Here is the notebook I used to test this: https://github.com/vandalt/jaxoplanet-astrometry/blob/main/notebooks/exoplanet_astrometry_fitting.ipynb.

I'd be happy to contribute this or another relative astrometry example as a tutorial if you think it would be useful. I'm also planning to do an astrometry+RV example soon.

Cheers!

### DIFF
--- a/src/jaxoplanet/orbits/keplerian.py
+++ b/src/jaxoplanet/orbits/keplerian.py
@@ -607,8 +607,7 @@ class OrbitalBody(eqx.Module):
             if parallax is None:
                 r0 = semimajor
             else:
-                # TODO
-                r0 = semimajor * constants.au * parallax
+                r0 = semimajor * parallax / constants.au
 
         sinf, cosf = self._get_true_anomaly(t)
         if self.eccentricity is None:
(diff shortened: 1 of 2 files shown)


soichiro-hattori:
Hi @vandalt! 
Thanks for the fix and new test! Happy to merge this now. 

I think testing the `jaxoplanet` astrometry features against `orbitize` would be a great idea as long as the dependencies work out nicely! 

There's also been some recent interest in using `jaxoplanet`'s astrometry features so we'd love for this notebook to be contributed as a tutorial!

PR #321, opened 2026-05-04
Units and consistency updates to positions and velocities for astrometry and RV

vandalt:
Following #319, I started working on a [combined RV + relative astrometry](https://github.com/vandalt/jaxoplanet-astrometry/blob/main/notebooks/bpic_astrometry_rv.ipynb) example. And encountered some issue. Instead of working around them in the notebook, I tried to implement fixes here directly.

Here is the list of changes:

- **Fix** the arcsec conversion for velocity by dividing by `constants.au` like #319 did for position.
- **Fix** the position and velocity docstrings to mention that arcseconds are used when the parallax is provided or set at the object-level.
- **Modify** the `radial_velocity` function to return the RVs in units of m/s (**breaking change**)
- **Add** `M_jup` (in units of `M_sun`) and (`R_sun_per_d` in units of `m/s`) to `constants.py`
- **Add** a system with a parallax to `tests/orbits/keplerian_test.py` to make sure all tests are passing with a non-None parallax as well.
- **Add** unit tests to check the consistency with commonly used packages ([RadVel](https://radvel.readthedocs.io/en/latest/) for RVs and [orbitize!](https://orbitize.readthedocs.io/en/latest/index.html) for astrometry)
- **Update** the unit tests to account for the changes above

I think the `radial_velocity` change is worth discussion here:

- Pro: RVs are pretty much always returned in m/s or km/s and the Rsun/d will be quite surprising for a lot of users.
- Pro: If the change is well documented in the `radial_velocity` docstring and limited to that function (and does not affect the `_get_position_and_velocity()`, I don't think it will be very disruptive for existing users.
- Con: It breaks consistency with the [units convention](https://jax.exoplanet.codes/en/latest/conventions/#units)
- Con: It can be worked around fairly easily by users, especially with the new `constants.Rsun_per_day`

- [ ] If we pick this option, we would need to update the docs to make note of this and change units in the RV tutorial.

Alternative solution: add warning banners to tutorials using RV data to outline the surprising units.

Happy to implement the alternative solution instead if you think it is better.

Also, once this PR is merged, I will make a PR with both notebooks from https://github.com/vandalt/jaxoplanet-astrometry to demonstrate astrometry alone and joint astrometry+RV if that works for you.

Thanks!

EDIT: Added a note about the required docs update.

### DIFF
--- a/docs/tutorials/quickstart.ipynb
+++ b/docs/tutorials/quickstart.ipynb
@@ -123,14 +123,12 @@
    "id": "10",
    "metadata": {},
    "source": [
-    "## Radial velocity"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "11",
-   "metadata": {},
-   "source": [
+    "## Radial velocity\n",
+    "\n",
+    "```{important}\n",
+    "Note that the radial velocity is returned in ``Rsun/day`` to follow the *jaxoplanet* [units convention](../conventions.ipynb). See the [Radial Velocities tutorial](../rv.ipynb) for more on this.\n",
+    "```\n",
+    "\n",
     "Then, one can access the relative position and velocity of the planet relative to the sun. "
    ]
   },
--- a/docs/tutorials/quickstart.ipynb
+++ b/docs/tutorials/quickstart.ipynb
@@ -135,7 +135,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "12",
+   "id": "11",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -150,7 +150,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "13",
+   "id": "12",
    "metadata": {},
    "source": [
     "```{note}\n",
@@ -163,7 +163,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "14",
+   "id": "13",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -187,7 +187,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "15",
+   "id": "14",
    "metadata": {},
    "source": [
     "## Occultation light curve of limb darkened star\n",
@@ -200,7 +200,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "16",
+   "id": "15",
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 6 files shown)


soichiro-hattori:
@vandalt I'm sorry I haven't gotten back about this PR! 
And thank you so much for working on this!

While I understand that Rsun/day is definitely a surprising unit, I'm still currently leaning towards `jaxoplanet` keeping an internally consistent unit system (i.e., `radial_velocity` still returning values in Rsun/day) as opposed to individual functions returning values in different units. 

So my vote would be for the alternative solution you mentioned where in the tutorials we make it clear that we use these units and quickly show how to convert to m/s or km/s.

I'm certainly open to discussing this further, but maybe we should open a new issue for that since it'd be great to merge the other changes you've made! 

Thanks again for working on this!

vandalt:
Thanks for the feedback @soichiro-hattori! I agree it makes sense to keep internally consistent units.

I will revert to internal code to Rsun/day everywhere and update the docs accordingly.

Thanks!

vandalt:
@soichiro-hattori I updated the RV tutorial in the docs and reverted the `radial_velocity` method to return Rsun/d now. I think the PR is ready for review.

Thanks!

soichiro-hattori:
@vandalt thanks!

I have a bit of a pedantic point I'd want your opinion on before we merge this!
To me it seems like with the current way we're naming things (`constants.m_per_s`), this constant is something that should be *multiplied* to a numerical value instead of *dividing* with it (since dividing with it would imply the resulting value implicitly has units of `s/m`)? 

What do you think about storing the reciprocal of the current value in `constants.m_per_s` and then multiplying it to the values? Very happy to hear your opinion or other better ideas!

vandalt:
@soichiro-hattori it seems that currently everything in `jaxoplanet.constants` is setup so that multiplying a value _in that unit_ by the constant will convert it to a value _in jaxoplanet default units_. For example `1 * constants.M_earth` will give one Mearth in Msun. Same for `1 constants.au` which will give one AU in Rsun.

As it is written right now, `1 * constants.m_per_s` will convert one m/s to units of Rsun/day, so I feel like this is the most consistent behaviour.

Let me know if that works or if I should switch to something else.

Thanks!

soichiro-hattori:
@vandalt that's fair! 
And if some users find this confusing later on maybe we'll consider renaming some of our constants!

Thanks again for your work on this! I'll merge this now.

PR #325, opened 2026-05-05
Support arviz>=1.0 in the docs

### LINKED ISSUE #323: `arviz>1.0` migration
Hi!

I've been using `arviz>1.0` in some projects recently and it seems to be working well.

Would you be open to a PR that updates the docs to versions 1.0 and above?

vandalt:
This adds support for arviz 1.0 in all existing tutorials. The changes are fairly minimal:

- `az.from_dict()` now wants an extra level mapping the group name. For prior samples I felt like `prior` was the best choice. Then we need to pass it to corner
- `backend_kwargs` now needs to be passed via `figure_kwargs`

Cheers!

Fixes #323.


vandalt:
Ah I had forgotten that I was was using the `main` branch of `corner` locally. This explains the `docs-build` failure.

Options would be:

1. Use the main branch of corner through a `git` dependency (a bit flaky and would require users to install this as well to follow the tutorials)
2. Wait for a new release of corner to use the latest arviz (Cc-ing @dfm in case he has the time to look into this)
3. Use something other than corner, even temporarily (e.g. `arviz.plot_pair`)

Option 2 is probably the best here. Also, nothing prevents users from installing arviz 1.0 if they want to since it is only a `docs` optional dependency.

vandalt:
Relevant issue upstream in corner: https://github.com/dfm/corner.py/issues/331

vandalt:
@soichiro-hattori I think this is ready for review now that a new corner version has been released and the docs are now building :)

PR #328, opened 2026-05-20
Joint astrometry and RV tutorial

vandalt:
This tutorial demonstrates how to build and fit a joint RV and relative astrometry model.
I used beta Pic since I worked on this a few years ago and wanted to try the model in JAX anyway.

The notebook takes quite some time to run since I explore a few cases:

1. Planet b only
2. Planet b+c
3. Planet b+c including a GRAVITY data point with a correlated 2D likelihood
4. Case 3 but with error inflation terms

This may be more appropriate as a case study but I'm opening a PR in case you want to merge it as a tutorial.

Thanks!

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -24,6 +24,7 @@ tutorials/ttv.ipynb
 tutorials/rv.ipynb
 tutorials/rossiter.ipynb
 tutorials/relative-astrometry.ipynb
+tutorials/relative-astrometry-rv-bpic.ipynb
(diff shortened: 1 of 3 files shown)


vandalt:
Trying to build on readthedocs results in a timeout. Should I configure this notebook to be executed locally? Or Remove the "extra parts" and only do a single MCMC run? Or just wait until there is a Case studies page?

Thanks!

